PR #1, opened 2019-05-07
README

joshtemple:
Wrote an initial draft of the README with motivation and a short description. This probably isn't a 1:1 with the abstract, but it's a start on the problem statement and initial design.


DylanBaker:
@joshtemple This looks good to me. Do you want to merge as-is and then we can add to it?

PR #4, opened 2019-05-24
Join 2019 Submission

DylanBaker:
Thought it might be easier to edit a file so moved [this issue](https://github.com/dbanalyticsco/Fonz/issues/3) across to this PR. 

(It's possible there is a way to convert the issue directly? I'm not super familiar with the workflow.)

### DIFF
--- a/docs/join2019_submisson.md
+++ b/docs/join2019_submisson.md
@@ -6,9 +6,11 @@
 
 Occasionally, when we make changes to LookML or transform an underlying database table, we break downstream experiences in Looker by introducing database or content errors. These bugs are easy to deploy to production unknowingly, since they are runtime errors, not compiler errors. In short, they aren't discovered until a user runs a query. For little-used fields, these bugs can remain hidden for weeks or months.
 
-We wanted a tool that would check for fields with database errors and content with LookML errors. We could run the tool in a continuous integration (CI) pipeline before deploying to production or after database changes. This would help establish a baseline performance expectation for our Looker instance. We believe in the power of CI for analytics, so we built a new tool, Fonz, to enhance the business intelligence layer of analytics CI pipelines.
+We believe in the power of CI for analytics, firmly of the view that analytics can adopt best practices from software engineering. We wanted a tool that would check for fields with database errors and content with LookML errors. We wanted to leverage continuous integration (CI) pipelines as part of LookML deployments or proposed database changes to provide a seamless development experience. Ultimately, we wanted to establish a baseline performance expectation for our Looker instance. 
 
-Fonz is built on the Looker API and automatically runs queries across enabled explores to check for database errors. Next, it performs content validation to catch any content that newly introduced code may have broken. Finally, Fonz can be run in conjunction with database code changes to catch cases where transformations will break a downstream Looker view. We believe Fonz is a valuable addition to the "Built on Looker" open-source ecosystem and will improve quality and reliability within Looker.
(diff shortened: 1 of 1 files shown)


docs/join2019_submisson.md
```diff
@@ -0,0 +1,29 @@
+#### Session Title
+
+### A Reliable Development Workflow with CI for Looker
```
DylanBaker: I'm trying to figure out what angle is most sexy for this type of tool. 

What do you think of something like: `Develop LookML Like a Software Engineer: CI for Looker". I don't think the title is likely to make a huge difference, but I think we can probably make it pretty punchy.

DylanBaker: Or even making it explicitly about the tool (which has a recognisable name)? 

"Introducing Fonz: a CI tool for reliable LookML and warehouse development"

docs/join2019_submisson.md
```diff
@@ -0,0 +1,31 @@
...
+
+#### Session Abstract
+
+Occasionally, when we make changes to LookML or transform an underlying database table, we break downstream experiences in Looker by introducing database or content errors. These bugs are easy to deploy to production unknowingly, since they are runtime errors, not compiler errors. In short, they aren't discovered until a user runs a query. For little-used fields, these bugs can remain hidden for weeks or months.
+
+We believe in the power of CI for analytics, firmly of the view that analytics can adopt best practices from software engineering. We wanted a tool that would check for fields with database errors and content with LookML errors. We wanted to leverage continuous integration (CI) pipelines as part of LookML deployments or proposed database changes to provide a seamless development experience. Ultimately, we wanted to establish a baseline performance expectation for our Looker instance.
```
DylanBaker: I've altered this slightly, making the second paragraph more explicitly about the values that drove the development of the tool. Let me know if you think that works. Happy to revert.

DylanBaker:
(Also, definitely don't have to merge this in. Just thought it would be the easiest way to comment.)

docs/join2019_submisson.md
joshtemple: I like both of those a lot better. I kept thinking... "How do we _actually_ get people excited about attending a session on CI?"

I think the first one is slightly better. It feels a bit more provocative to me.

docs/join2019_submisson.md
```diff
@@ -0,0 +1,31 @@
+#### Session Title
+
+### Develop LookML Like a Software Engineer: Continuous Integration for Looker
+
+#### Session Abstract
+
+Occasionally, when we make changes to LookML or transform an underlying database table, we break downstream experiences in Looker by introducing database or content errors. These bugs are easy to deploy to production unknowingly, since they are runtime errors, not compiler errors. In short, they aren't discovered until a user runs a query. For little-used fields, these bugs can remain hidden for weeks or months.
+
+We wanted a tool that would check for fields with database errors and content with LookML errors. We firmly believe that analytics should adopt best practices from software engineering, and we're especially excited by the power of continuous integration (CI). We could run this tool in a CI pipeline before deploying to production or after database changes, helping to establish a baseline performance expectation for our Looker instance.
```
joshtemple: I adjusted this paragraph slightly to make it flow a bit more clearly.

joshtemple:
Updated the title and made a small change in the second paragraph. If it looks good to you, merge away! This closes #4.

DylanBaker:
@joshtemple That's merged in. Do you want to send the proposal itself? I'm also happy to do so.

PR #25, opened 2019-06-08
Exception handling

### LINKED ISSUE #18: Top-level exception handler
We need to build some error handling into Fonz, starting with an overall exception handler that catches exceptions and prints their results rather than letting exceptions propagate through the stack to `stderr`.

Here's how I would recommend doing this.

1. We create a base exception class `FonzException` inheriting from `Exception`. This is the exception class we use to throw our own errors (bad credentials, missing branch, etc.). We can subclass `FonzException` if we want more granularity on the types of exceptions we throw.

1. We implement a method `FonzException.show` which logs any message we specify in the exception to the command line. For example, we might catch a `requests.exceptions.HttpError` and re-raise it as a `FonzException` with a more helpful message "Looker base URL does not exist, please try again."

1. In our top level of execution (`cli.py:cli`), we add a try/catch block (I'm not 100% of how this will interact with Click, but this is how I would do this normally). The try catch block first attempts to catch `FonzException`. If it catches one, it calls the `show` method and exits with a nonzero status code.

1. The next `except` block catches `Exception`, so basically anything unhandled or unexpected. It logs the error message (but not the full traceback) to the command line as a generic runtime error. It also exits with a nonzero status code.

1. If we want to go farther, we can log the full traceback to some kind of log file, similar to how dbt does this.

The general principle is that all exceptions are caught at the top level of execution (via Python's Exception base class) and handled by logging output to the command line and exiting with a nonzero code, rather than raising the exception all the way out, resulting in a large Python traceback.

joshtemple:
Closes #18.

### Fonz API changes

The main loop still lives in `cli.sql`, where we use the `Fonz` instance to collect the explores, get their dimensions, and validate each one. Finally, the main loop is responsible for printing any results.

Throughout the code, I've **maintained a naming convention** where `explore` refers to the JSON dictionary and `explore_name` is the string version.

I've **removed the concept of `explore['failed']`**, replacing it with an exception handling approach to raise and handle `SqlError` exceptions as follows:

### Exception handling for SQL errors

If an API call returns a Looker error:
1. Collect the error message
1. Raise a custom `SqlError` exception with the explore name, query ID, and error message

To handle the `SqlError` exception:
1. Collect the query's raw SQL and log it to a file
1. Try to parse the line number from the query and collect the context of the SQL being executed (within some window of lines)
1. Save the SQL error message and SQL context to `Fonz.messages` instance variable to be printed at the end

### Additional exception handling

I created a top-level exception handler in `cli.py` to catch anything that flows up and print it nicely, logging the traceback at the debug level so it's saved to file.

I had to use this exception handler as a decorator in the extended `click.Command` class because click didn't like me putting other decorators on top of the `connect` and `sql` methods. The general idea is outlined in #18 and is to catch things we raise specifically as `FonzExceptions` or its children and handle them slightly differently than children of `Exception` (basically just unexpected, unhandled errors that are thrown). 

Throughout the code, we can catch certain errors and re-raise them as `FonzExceptions` with additional context when we have a good guess of what happened (e.g. a bad authentication).

### Minor changes

I'm trying to use [f-strings](https://realpython.com/python-f-strings/) instead of `format` these days, so I replaced a few of those string formattings for readability.

I removed `query_url` since I didn't think it added enough additional info to be worth the complexity of implementation. What I think would be _more_ useful would be to link to the LookML, similar to how hovering over a dimension provides the Looker logo that links directly to the dimension definition. I'm not sure where that link comes from or if it's exposed in the API but that would be super helpful.


joshtemple:
I fixed up all the CLI tests, which were failing locally because pytest wasn't running with clean/empty environment variables and was using my current environment. That should be fixed now.

It looks like there's still one failing test in `test_connection`, which seems to be due to the fact that the `dimensions` object created in the mock doesn't have a `sql` key. @DylanBaker, are you able to get that one passing?

Leaving this PR in WIP state for now until we agree on this approach and I can write some more tests.

joshtemple:
@DylanBaker, I believe this does what we need it to for now and can be merged in (once we get those tests from #26 merged) once you review.

PR #26, opened 2019-06-09
Testing connection.py

DylanBaker:


fonz/connection.py
```diff
@@ -39,7 +39,7 @@ def connect(self) -> None:
 
         response = self.session.post(
             url=compose_url(self.base_url, path=["login"]),
-            data={"client_id": self.client_id, "client_secret": self.client_secret},
+            json={"client_id": self.client_id, "client_secret": self.client_secret},
```
joshtemple: Any specific reason why you're using `json` here? The example in the API docs uses `Content-Type: application/x-www-form-urlencoded` which I believe corresponds to the `data` argument to `post`.

joshtemple: API docs for reference: https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/api-auth

PR #38, opened 2019-06-15
Click to argparse

DylanBaker:


DylanBaker:
@joshtemple Have had a first go at refactoring cli.py to use argparse. I'm pretty sure the code itself works, but I need to get my head around the required changes to the testing, which I'll do next.

Issue #23, opened 2019-06-08
Provide SQL query exerpt as part of error message


DylanBaker:
There are a number of instances where the error message passed through from Looker isn't that helpful in debugging what needs to be changed, i.e. `Syntax error: Unexpected ":" at [574:12]"`. When this happens, I have to jump into the `logs/filename.sql` to find the line of SQL that is being referred to in the error. 

Ideally, it would be easier to see right away. We should be able to parse the line reference in the error message and print the section of the SQL query itself that is erroring.

PR #39, opened 2019-06-17
Query dimensions individually

### LINKED ISSUE #20: Asynchronously query explores
To speed up Fonz, we should consider using the `Run Query Async` endpoint and parallelizing our requests to run each explore. We can essentially launch each query required, then check on the status of each until a result is obtained. We should probably be able to set a limit on the number of parallel queries in case a database has low concurrency limits.

I'm guessing the `asyncio` library also has some useful tools for this.

If we get enough of a speed up, we can consider a toggled mode (or perhaps even default behavior) to do one query per dimension. This improves the UX because you can see all existing errors at once, rather than one error per explore.

joshtemple:
Closes #20, #23, #42.

### Project LookML representation

Added a new module, `lookml.py`, which contains a few classes to represent LookML objects like explores, models, and dimensions. These classes can be constructed alternatively via their `from_json` methods, which directly parse the API JSON responses (supplied in `dict` form).

Added a new method, `Fonz.build_project`, which builds the entire tree, including models, explores, and dimensions. This tree is stored in the attribute, `lookml` and is of type `Project`.

All of these objects, except `Project`, have a boolean `errored` attribute, which is set when an error is returned by a query. This attribute is used in the reporting step at the end to determine which objects returned query errors. Explores and Dimensions have an `error` attribute, which can hold a `SqlError` instance containing the queried SQL, the line number, and error message returned by the API.

### Simplifying `cli.py`

I've moved the main loop out of `cli.py` into `Fonz.validate` and `Fonz.report_results`. The new flow is as follows in `cli.py` and is much cleaner.

```
client.connect()
client.update_session()
client.build_project()
client.validate(batch)
client.report_results(batch)
```

### Batch option

You can now provide `--batch` or `--no-batch` to Fonz via the command line. The default, when nothing is specified, is to run without batching. Batch mode runs one query per explore, while no-batch mode runs one query per dimension.

### Asynchronous querying for dimensions and explores

All querying of the Looker Query endpoint uses the Async Query approach instead of Run Query. Since the queries are run asynchronously, this approach adds an extra step: create > run > check back for results until done. The run step returns a query task ID which can be used to check for results periodically.

Luckily, the Async Query endpoint is also supported in API v3.0, although it is in beta status for both 3.0 and 3.1 and is subject to changes. This endpoint is great because it returns the SQL and line number of the error already parsed in the response when there's an error.

I'm using the `backoff` library to exponentially slowdown checking for results if the query results are still pending. For example, the check results step could wait 0.1, 0.5, 1, 2, and 4 seconds sequentially until results are obtained. This limits the overall load of requests against the endpoint. We should tweak the curve used to backoff for performance tuning down the road, I haven't changed it much so far.

Finally, I'm leveraging `asyncio` and `aiohttp` libraries to run the requests asynchronously. This means that while we're waiting for the response back from Looker's API, we can launch other requests. This adds some complexity to the code, but it's a big speedup.

### Other changes

When running in no-batch mode, Fonz will show the link to the LookML of each errored dimension.

`Fonz.base_url` is now `Fonz.api_url` because we need to use the actual base URL (minus the API stuff) to build the LookML link.

Logging is slightly more sophisticated, generating subfolder trees for models and explores (if in no-batch mode).

Removed `tests/mock.py` in favor of a standard library approach which is more compatible/consistent with approaches for testing the async functions. Some of the async testing is really hard to follow, so happy to explain where necessary.

Some tests load fake JSON API response files from `tests/resources`. I have a useful convenience method called `load` that does this more easily. Feel free to take advantage of that.


.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           name: mypy
           command: |
             . venv/bin/activate
-            mypy fonz
+            mypy fonz --ignore-missing-imports
```
joshtemple: mypy keeps throwing an import error for a package that is definitely imported and is working as far as I can tell. I confirmed that the package is installed in our CI build so I'm just disabling this feature because I'm not sure what else to do about it.

DylanBaker: Cool. Happy to disable for the moment. We can always come back to it in the future.

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Dylan Baker
+Copyright (c) 2019 Dylan Baker & Josh Temple
```
DylanBaker: 👍 🔥

fonz/connection.py
```diff
@@ -146,100 +255,121 @@ def get_dimensions(self, model: str, explore_name: str) -> List[str]:
...
+            result = await self.get_query_results(
+                async_session, query_task_id, explore.name
             )
-        sql = response.text
 
-        return sql
+            if not result:
+                return
+            elif isinstance(result, dict) and result.get("errors"):
+                first_error = result["errors"][0]
```
DylanBaker: Is this `first_error` in that there are multiple errors or `first_error` in that the error in in the first 'row'?

fonz/lookml.py
```diff
@@ -0,0 +1,122 @@
...
+        return f"{self.__class__.__name__}(name={self.name})"
+
+
+class Dimension(LookMlObject):
+    def __init__(self, name: str, type: str, sql: str, url: str):
+        self.name = name
+        self.type = type
+        self.sql = sql
+        self.url = url
+        self.ignore = True if "fonz: ignore" in sql else False
```
DylanBaker: Is it currently implemented to actually ignore based on this?

fonz/lookml.py
```diff
@@ -0,0 +1,122 @@
+from typing import List, Optional
+from fonz.exceptions import SqlError
+
+
+class LookMlObject:
+    def __repr__(self):
+        return f"{self.__class__.__name__}(name={self.name})"
+
+
+class Dimension(LookMlObject):
```
DylanBaker: Should the `Dimension` class be aware of what explore it is in or do we think that it will always be implemented as part of the `Explore` class?

fonz/lookml.py
```diff
@@ -0,0 +1,122 @@
...
+    @classmethod
+    def from_json(cls, json_dict):
+        name = json_dict["name"]
+        type = json_dict["type"]
+        sql = json_dict["sql"]
+        url = json_dict["lookml_link"]
+        return cls(name, type, sql, url)
+
+
+class Explore(LookMlObject):
```
DylanBaker: Likewise, should an explore know what model it is from?

fonz/lookml.py
joshtemple: I thought about this and did some looking into how tree structures are implemented. It turns out that having a node that references both its children and its parents is not recommended (basically it's no longer a directed graph since you can traverse both directions). Let's leave it this way until we encounter a need that indicates otherwise.

fonz/connection.py
joshtemple: It's the first error in the array of errors returned by the API. The API response for the Async Query Result endpoint seems a little half-baked, but it seems you can get multiple errors back from the endpoint.

fonz/lookml.py
joshtemple: Not yet!

Issue #34, opened 2019-06-12
SQL returned by the Looker API differs from executed


joshtemple:
I've identified an issue with Looker where the SQL query string returned by the Query API is not the same as the query actually executed against the data warehouse.

Looker prepends a comment in the first line, and I've identified cases where a few extra newlines are added in spots. This means that the error line number returned by the warehouse (and subsequently by Looker) doesn't match up with the error-causing line in the SQL returned by Looker.

Because of the arbitrary newlines, I can't account for this by assuming an extra line at the start, so this effectively breaks the ability to preview the SQL line that caused the error.

I'll update this issue once I have the bug reference number back from Looker.

DylanBaker:
Oof. That's annoying. Have they indicated that this is something they hope to resolve?

DylanBaker:
On Snowflake customers, I'm now seeing these comments at the bottom of the SQL, not the top. I've asked Looker if that's the case for all warehouses. If so, this might now be resolved.

PR #58, opened 2019-08-28
Misc features + model selection and dimension ignore

joshtemple:
This PR implements a bunch of tiny improvements plus some big ones. Test coverage is pretty underwhelming but I'd like to wait until this project is slightly more stable before I spend any more time writing tests.

### Reorganized architecture

`Runner` is the top-level API. Right now it only has one validate method (`validate_sql`), but eventually it will have validation methods for data tests, content validation, and/or linting. I have the runner set up to return dictionaries. These will be easy to turn into JSON in a Flask API via [`jsonify`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.json.jsonify).

These validate methods on the `Runner` object should create a instance of a `Validator` subclass (`validators.py`), which should have a `validate` method. Right now, it's just `SqlValidator`. The validator is responsible for handling the bulk of the logic around what is valid and what is not.

The validator itself should never make HTTP requests. Instead, the validator interacts with a `LookerClient` (`client.py`) instance, which is basically a wrapper around the API requests we care about.

The main thing to wrap your head around here is that the Validator should primarily be responsible for returning results of validation in a dictionary format and should be UI-agnostic. Most of the logging/printing that was previously happening here has been moved to `cli.py`.

### Multi-results endpoint

Previously, we were running async requests to create a query, create a query task, then ping the results endpoint repeatedly for that task until a result was ready.

The first two steps are basically the same: asynchronously create queries and query tasks. However, results checking is now a synchronous step that pings the multi results endpoint every half second until all queries are complete and results have been obtained.

This _should_ be faster and more resource efficient since we won't have to generate a separate get results request for every dimension or explore.

### Notable other changes

* Better UI and printing all around
* `--explores` argument for model selection (supports `*` wildcard)
* Allow users to ignore dimensions by adding `-- fonz: ignore` to the dimension's SQL
* Respect `NO_COLOR` env var and `TERM=dumb` and don't print color codes in those situations
* Prevent color codes from printing to file via a custom logging formatter
* Lots of docstrings (not 100% but getting there)
* Introduce `flake8` into CI suite for extra checking
* Fixed precedence in the CLI by adding some custom actions for parsing config files and environment variables. This was a huge pain to build and is a bit hackier than I would like. It seems to work as intended though according to my tests.
* Somewhere along the way we lost the ability to log SQL queries to file. That will have to be added back in.
* I disabled the SQL context in the error message until #34 is resolved.
* Updated the way the `errored` attribute works on LookMlObjects. Now it propagates up and down the tree automatically (e.g. a model will return `errored = True` as soon as a dimension in one of its explores as an error set to its `error` attribute.


joshtemple:
@DylanBaker, this could definitely use more testing, and I'll try and add some more tests on the model selection portion especially, but mind reviewing in the interim? This shouldn't be merged until after the argparse changes have been merged.

fonz/cli.py
```diff
@@ -11,92 +14,163 @@ def wrapper(*args, **kwargs):
         try:
             return function(*args, **kwargs)
         except ValidationError as error:
-            logger.error(f"{error}\n")
```
DylanBaker: What gets displayed when this error occurs?

fonz/connection.py
```diff
@@ -257,6 +328,9 @@ def get_dimensions(self, model: str, explore_name: str) -> List[str]:
 
         return response.json()["fields"]["dimensions"]
 
+    @backoff.on_exception(
```
DylanBaker: Is this to handle errors we are seeing from Looker?

fonz/cli.py
joshtemple: Instead it prints a [header line](https://github.com/dbanalyticsco/Fonz/blob/feature/misc/fonz/connection.py#L289-L296) with the SQL error in `connection.py` before the exception is raised. I could move those error parameters into the `ValidationError` object and handle the printing in `cli.py:wrapper` if you think that would be cleaner.

fonz/connection.py
joshtemple: Basically it's a retry in case we get a timeout or unexpected error.

joshtemple:
Phew! This is a lot! @DylanBaker, you're good to review this. I'm not sure how helpful the diffs will be given the magnitude of changes here. Let me know if you have questions or want to walk through specific sections.

PR #74, opened 2019-10-02
Documentation for running spectacles

DylanBaker:


DylanBaker:
@joshtemple still a WIP, but thought I'd push here for discussion. 

My current thinking for structure is:

* Why we built this
* Installation
* Getting Started
  * Looker user
  * Required Arguments
* Testing your connection
* SQL validator
  * Additional Arguments

Any strong feelings?

joshtemple:
Your outline looks right to me. I would change "SQL validator" to a larger section called "Validators" where "SQL validator" is a sub-section of that. That section should also mention the `--spectacles: ignore` syntax.

The "Required arguments" section I would rename to "Configuration" and talk about all your options for specifying.

Somewhere in here we should highlight the state issue, maybe in the section on "Looker user"?

Should we close this PR in favor of [spectacles-docs](https://github.com/spectacles-ci/spectacles-docs)?

DylanBaker:
My worry about putting validators all in one section is that it becomes a very long page on the docs site at the point where we have more than two of them. That's a problem for the future though and I think the setup you suggest makes sense for the moment. 

Closing this and will open PR on spectacles-docs.

PR #76, opened 2019-10-15
Improve error messaging for API errors

joshtemple:
Include the `message` key from the response JSON, if found, in `APIConnectionError` messaging. Add a utility to pull that string out of the response object. Also add suggestion for branch checkout errors to ensure there are no uncommitted changes.


joshtemple:
Great! I had a similar thought but also couldn't come up with a better way to abstract beyond what I have here.

PR #77, opened 2019-10-16
Add data testing validator

### LINKED ISSUE #62: Trigger Looker data tests
We should support calling the API endpoint for Looker data tests when the feature launches. I imagine this should be a separate subcommand (`sql`, `connect`) which will run the Looker data tests and return a result. I'm not sure what the appropriate subcommand will be.

I imagine the final workflow for this tool being something like (all steps optional):

1. run a LookML linter
1. run SQL dimension tests
1. run all Looker data tests (these seem like they would be more complex assertions)
1. run content validation

joshtemple:
Adds a new validator with command `spectacles assert` to support running LookML/data tests via the API.

The following outstanding items need to be completed before this is ready for review:
- [ ] Looker pushes a fix to include more metadata (explore, line number, etc.)
- [ ] Looker pushes a fix to include the name of the assertion in the API response
- [ ] Flesh out error message with above metadata
- [ ] Provide pass/fail at the explore level for all explores affected by data tests
- [ ] Write unit tests

Closes #62.

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -156,7 +156,7 @@ def main():
             handler.setLevel(args.log_level)
 
     if args.command == "connect":
-        connect(
+        run_connect(
             args.base_url,
             args.client_id,
             args.client_secret,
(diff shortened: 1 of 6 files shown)


joshtemple:
Going to merge this in for now, can reopen and push updates as Looker releases these API changes.

PR #78, opened 2019-10-16
Workaround for 500 errors

### LINKED ISSUE #73: 500 errors for get_query_task_multi_results
We already have a ticket open with Looker (ticket #: EMBED-466) for this, but adding here to document.

After a few queries to `get_query_task_multi_results`, we receive a 500 Internal Server Error. Looker support originally believed this was because of a cache miss, but we still get this error running with `cache=false` as a query string or in the JSON body of the `create_query_task` request.

joshtemple:
Switches queries from result format of `json` to `json_detail` per Looker's advice.

Turns SSL validation back on, which was previously deactivated for testing.

Closes #73.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -193,7 +193,7 @@ class SqlValidator(Validator):
         loop.run_until_complete(asyncio.sleep(0.250))
         loop.close()
 
-        MAX_QUERY_FETCH = 292
+        MAX_QUERY_FETCH = 250
 
         tasks_to_check = query_task_ids[:MAX_QUERY_FETCH]
         del query_task_ids[:MAX_QUERY_FETCH]
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -170,20 +170,28 @@ class SqlValidator(Validator):
         )
 
         loop = asyncio.get_event_loop()
+        session = aiohttp.ClientSession(
+            loop=loop, headers=self.client.session.headers, timeout=self.timeout
+        )
         tasks = []
         for model in self.project.models:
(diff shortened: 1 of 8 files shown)


spectacles/validators.py
```diff
@@ -193,11 +193,21 @@ def validate(self, batch: bool = False) -> List[SqlError]:
         loop.run_until_complete(asyncio.sleep(0.250))
         loop.close()
 
-        running_task_ids, errors = self._get_query_results(query_task_ids)
-        while running_task_ids:
-            running_task_ids, more_errors = self._get_query_results(running_task_ids)
+        MAX_QUERY_FETCH = 292
```
joshtemple: Let's round this down to 250 just to be safe.

spectacles/validators.py
```diff
@@ -193,11 +193,21 @@ def validate(self, batch: bool = False) -> List[SqlError]:
         loop.run_until_complete(asyncio.sleep(0.250))
         loop.close()
 
-        running_task_ids, errors = self._get_query_results(query_task_ids)
-        while running_task_ids:
```
joshtemple: I wonder if it makes it would be simpler to put this `while` in a `for` loop that iterates over the list in batches.

You could use something like [this](https://stackoverflow.com/a/312464/3524951). Thoughts?

DylanBaker: I think it would be less efficient. You'd have to wait for a full chunk to finish before kicking off the next one, wouldn't you? This way, we are always querying the max amount. (Let me know if I misunderstood the recommendation.)

PR #87, opened 2019-10-30
Validate instance version before SQL validation

### LINKED ISSUE #81: Validate and display Looker instance version
Since spectacles now requires a certain version of Looker to run successfully, we should output the version of the user's Looker instance and validate that it's >= 6.20.12. This is especially needed for users with self-hosted instances which likely will not be as up to date.

joshtemple:
Closes #81.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -43,9 +43,18 @@ class SqlValidator(Validator):
     """
 
     timeout = aiohttp.ClientTimeout(total=300)
+    MIN_LOOKER_VERSION = "6.22.12"
 
     def __init__(self, client: LookerClient, project: str):
         super().__init__(client)
+        meets_required_version = self.client.validate_looker_release_version(
+            required_version=self.MIN_LOOKER_VERSION
+        )
+        if not meets_required_version:
+            raise SpectaclesException(
+                "SQL validation requires version "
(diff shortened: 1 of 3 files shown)


spectacles/validators.py
```diff
@@ -43,9 +43,18 @@ class SqlValidator(Validator):
     """
 
     timeout = aiohttp.ClientTimeout(total=300)
+    MIN_LOOKER_VERSION = "6.22.12"
```
agile: Just wanting to confirm this is in fact the min version required, because #81 mentions `6.20.12`?

joshtemple:
Looks like the version in #81 is a typo. However, we were told by Looker
that the API fix required here should be backported to 6.20, but I’m not
sure if that’s happened yet.

Anything after 6.22.12 should have the fix we’re checking for here, but if
the back port has occurred it should work on 6.20+

If you are able to get it working on pre-6.22.12, please let us know!

spectacles/validators.py
joshtemple: Looks like the version in #81 is a typo (I've fixed it). However, we were told by Looker
that the API fix required here should be backported to 6.20, but I’m not
sure if that’s happened yet.

Anything after 6.22.12 should have the fix we’re checking for here, but if
the back port has occurred it should work on 6.20+

If you are able to get it working on pre-6.22.12, please let us know!

PR #95, opened 2019-11-04
Add remote reset flag

### LINKED ISSUE #89: Pull updated changes from remote after checkout
Currently, spectacles does not pull changes from remote if the local branch is behind the remote branch. This is a problem if the user making changes to the branch is different than the user running spectacles (e.g. a bot user set up for the sole purpose of running spectacles queries). In that case, the spectacles bot user would not have the most recent changes unless it had some way of pulling them from remote.

There is a way to [reset to remote via the API](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/project#reset_to_remote), but will delete any saved or committed changes

However, spectacles can determine the [number of commits](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/project#get_active_git_branch) it is ahead and behind the remote. We may be able to do some kind of checking to anticipate if commits would be deleted and fail with an error message or something. This would avoid resetting commits not pushed to remote, but probably wouldn't work for things that are saved by not committed.

DylanBaker:
Closes #89 

@joshtemple Wanted to test out one implementation of this. As discussed, happy to go another route if preferred.

Adds a flag `--remote-reset` to the SQL validator which tells spectacles to reset the branch to the remote version after checking it out.

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -174,6 +174,7 @@ def main():
             args.port,
             args.api_version,
             args.mode,
+            args.remote_reset,
         )
 
 
@@ -333,6 +334,13 @@ def _build_sql_subparser(
             query per explore. In hybrid mode, the SQL validator will run in \
             batch mode and then run errored explores in single-dimension mode.",
     )
+    subparser.add_argument(
+        "--remote-reset",
+        action="store_true",
+        help="When set to true, the SQL validator will tell Looker to reset the \
+            user's branch to the revision of the branch that is on the remote. \
+            This will delete any uncommited changes in the user's workspace.",
+    )
 
 
 def connect(
@@ -352,10 +360,18 @@ def sql(
     port,
     api_version,
     mode,
+    remote_reset,
 ) -> None:
     """Runs and validates the SQL for each selected LookML dimension."""
     runner = Runner(
-        base_url, project, branch, client_id, client_secret, port, api_version
+        base_url,
+        project,
+        branch,
+        client_id,
+        client_secret,
+        port,
+        api_version,
+        remote_reset,
     )
     errors = runner.validate_sql(explores, mode)
     if errors:
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -339,7 +339,7 @@ def _build_sql_subparser(
         action="store_true",
         help="When set to true, the SQL validator will tell Looker to reset the \
             user's branch to the revision of the branch that is on the remote. \
-            This will delete any uncommited changes in the user's workspace.",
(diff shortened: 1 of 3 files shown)


iserko:
please merge this 🙏  I spent the last 2 hours figuring out why my deleted explore/view was still being read by Spectacles. I'm using this code in my local runs and it worked nicely

DylanBaker:
@joshtemple Did you want to consider a different name for this flag, or happy to merge in as is?

DylanBaker:
> please merge this 🙏 I spent the last 2 hours figuring out why my deleted explore/view was still being read by Spectacles. I'm using this code in my local runs and it worked nicely

We did exactly this before opening this PR. Will get this merged in shortly.

spectacles/cli.py
```diff
@@ -333,6 +334,13 @@ def _build_sql_subparser(
             query per explore. In hybrid mode, the SQL validator will run in \
             batch mode and then run errored explores in single-dimension mode.",
     )
+    subparser.add_argument(
+        "--remote-reset",
+        action="store_true",
+        help="When set to true, the SQL validator will tell Looker to reset the \
+            user's branch to the revision of the branch that is on the remote. \
+            This will delete any uncommited changes in the user's workspace.",
```
joshtemple: Can we add WARNING in all caps here or something to make this more obvious?

spectacles/cli.py
```diff
@@ -352,10 +360,18 @@ def sql(
...
     runner = Runner(
-        base_url, project, branch, client_id, client_secret, port, api_version
+        base_url,
+        project,
+        branch,
+        client_id,
+        client_secret,
+        port,
+        api_version,
+        remote_reset,
```
joshtemple: Since this is a SQL Validator-specific flag, instead of setting this in the constructor, this should be passed into `validate_sql` as an argument and used to instantiate the `SqlValidator`. Reason is we don't want to have to pass `remote_reset` for validators that are unrelated or don't care about this.

spectacles/runner.py
```diff
@@ -29,12 +29,13 @@ def __init__(
...
         port: int = 19999,
         api_version: float = 3.1,
+        remote_reset: bool = False,
     ):
         self.project = project
         self.client = LookerClient(
             base_url, client_id, client_secret, port, api_version
         )
-        self.client.update_session(project, branch)
+        self.client.update_session(project, branch, remote_reset)
```
joshtemple: In line with comments above, I would move this into `validate_sql`.

spectacles/cli.py
joshtemple: We decided to leave this in the runner since it _does_ apply to all validators. Thanks @DylanBaker!

Issue #99, opened 2019-11-08
Explores with 0 dimensions have IndexError


edmundyan:
Some of our explores have 0 dimensions (they might only have measures).  I get a runtime error when running in batch mode.

```
2019-11-07 20:06:06,433 DEBUG | list index out of range
Traceback (most recent call last):
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/cli.py", line 120, in wrapper
    return function(*args, **kwargs)
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/cli.py", line 176, in main
    args.mode,
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/cli.py", line 413, in run_sql
    errors = runner.validate_sql(explores, mode)
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/runner.py", line 42, in validate_sql
    errors = sql_validator.validate(mode)
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/validators.py", line 214, in validate
    errors = self._query(mode)
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/validators.py", line 254, in _query
    query_task_ids = list(loop.run_until_complete(asyncio.gather(*tasks)))
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 579, in run_until_complete
    return future.result()
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/validators.py", line 365, in _query_explore
    session, model.name, explore.name, dimensions
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/backoff/_async.py", line 131, in retry
    ret = await target(*args, **kwargs)
  File "/Users/eyan/.local/share/virtualenvs/looker-XwxA0sgW/lib/python3.7/site-packages/spectacles/client.py", line 352, in create_query
    dimensions[0]
IndexError: list index out of range
[...]

PR #101, opened 2019-11-08
Skip explores with no dimensions

edmundyan:
Fixes https://github.com/spectacles-ci/spectacles/issues/99

Honestly, a bit unfamiliar with the codebase, so feel free to re-implement how you'd like, but this did the trick for me locally.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -239,10 +239,11 @@ class SqlValidator(Validator):
             for explore in model.explores:
                 if mode == "batch" or (mode == "hybrid" and not explore.queried):
                     logger.debug("Querying one explore at at time")
-                    task = loop.create_task(
-                        self._query_explore(session, model, explore)
-                    )
-                    tasks.append(task)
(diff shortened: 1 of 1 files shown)


spectacles/validators.py
```diff
@@ -239,10 +239,11 @@ def _query(self, mode: str = "batch") -> List[SqlError]:
             for explore in model.explores:
                 if mode == "batch" or (mode == "hybrid" and not explore.queried):
                     logger.debug("Querying one explore at at time")
-                    task = loop.create_task(
-                        self._query_explore(session, model, explore)
-                    )
-                    tasks.append(task)
+                    if explore.dimensions:
```
joshtemple: Awesome. Could you bump this up a level to line 240? We want to catch this regardless of which query mode we're in and skip to the next explore in the model. That should do it.

joshtemple:
This looks great, except I think the indentation is off (2 instead of 4 spaces). Any idea what's going on there?

Issue #100, opened 2019-11-08
`_extract_error_details()` raises KeyError when there is no 'line' key


edmundyan:
In `_extract_error_details()`, the `sql_error_loc` dict does not always have a line number.  I suspect this is a Redshift-specific error messaging.

Here are some of the local vars of that function when I dig into the attrs of query_result.
```python
>>> message
'ERROR: syntax error at or near "AS"\n  Position: 6173'
>>> first_error['sql_error_loc']
{'character': 6173, 'message': 'syntax error'}
```

PR #102, opened 2019-11-08
Set line_number to None if key does not exist

edmundyan:
Fixes https://github.com/spectacles-ci/spectacles/issues/100

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -289,7 +289,7 @@ class SqlValidator(Validator):
                 )
             sql = data["sql"]
             if first_error.get("sql_error_loc"):
-                line_number = first_error["sql_error_loc"]["line"]
+                line_number = first_error["sql_error_loc"].get("line")
             else:
                 line_number = None
         elif isinstance(data, list):


edmundyan:
Are you asking because you want something similar to https://github.com/spectacles-ci/spectacles/blob/16d7f7aeecd9bb9f61196829929d8028e9292a2e/spectacles/printer.py#L68

I definitely think that would be useful, although I would suggest a more tighter coupling with the `SqlError` class.

But like you said `GET /query_tasks/{query_task_id}/results` is still a beta API so I'd be wary of building too much against that.  I'm also not sure how helpful "sql location context" is to me.  In the end, I would rather just know that it failed, a best attempt at error reason, and a link to the Explore so I can dig deeper myself.

joshtemple:
That makes sense. I agree that we're less concerned about the SQL context (which isn't even implemented yet because the Looker API doesn't return the correct line number, see #34) and more interested in the actual SQL error. I'm okay to merge this and think about ways to enhance the functionality down the road.

PR #104, opened 2019-11-11
Handle expired queries as errors

iserko:
On our Looker instance we had an issue with an explore
which took a long time to generate (over an hour actually),
and I had to kill the query. Killing the query makes the
query status 'expired', so this commit marks expired as
errors.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -319,10 +319,10 @@ class SqlValidator(Validator):
             query_status = query_result["status"]
             logger.debug("Query task %s status is %s", query_task_id, query_status)
 
-            if query_status in ("running", "added", "expired"):
+            if query_status in ("running", "added"):
                 still_running.append(query_task_id)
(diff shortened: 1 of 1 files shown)


joshtemple:
@iserko, thanks for reporting this. Your change is definitely the correct way to handle things, and was how we originally set things up.

We started seeing a strange issue where Looker's API would report that a query we had just launched as `expired` right before it completed successfully (as confirmed in the Looker Admin > Queries page). This would cause spectacles to incorrectly report the query as errored. I've created #111 for that problem, and we'll reach out to Looker to see if we can reproduce it and get a fix.

In the meantime, I'm going to leave this open (as I think the situation you're reporting is less common than the case I described), but will gladly merge this once #111 is closed.

iserko:
@joshtemple I just read the linked and will see if I can reproduce it on ours as well, but good catch 👍

PR #105, opened 2019-11-11
remove extraneous argument in the function docs

iserko:


joshtemple:
Thanks @iserko, good catch! 👍

PR #110, opened 2019-11-11
Support max_concurrency for running Looker queries

iserko:
Situation was that once we ran Spectacles on our Looker instance
with 128 explores in our model, it killed the instance in a way that
it stopped responding and eventually crashed. I've diagnosed that it
was merely the number of running queries at a time that caused it.
It scheduled pretty much all 128 queries immediately.

So this adds a flag --max-concurrency which you can use to tweak
and set how much load you want to put on your Looker instance.

I've tried our staging server with 10 and it works nicely and doesn't
crash anything.

Of the things I did:
- made the validate SQL function async and moved the loop code in the parent function
- got rid of creating an asyncio loop as I don’t believe you need it (get_event_loop creates it if one does not exist)
- when you specify max_concurrency the code will only create <max_concurrency> queries as will not queue up more until the previous ones are finished

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Sequence, DefaultDict, Set, Tuple
+from typing import List, Sequence, DefaultDict
 import asyncio
 from abc import ABC, abstractmethod
 from collections import defaultdict
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -80,12 +80,12 @@ class SqlValidator(Validator):
 
     timeout = aiohttp.ClientTimeout(total=300)
 
-    def __init__(self, client: LookerClient, project: str, query_slots: int = 10):
+    def __init__(self, client: LookerClient, project: str, concurrency: int = 10):
         super().__init__(client)
 
         self.project = Project(project, models=[])
         self.query_tasks: dict = {}
-        self.query_slots = asyncio.BoundedSemaphore(query_slots)
+        self.query_slots = asyncio.BoundedSemaphore(concurrency)
         self.running_query_tasks: asyncio.Queue = asyncio.Queue()
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


spectacles/validators.py
```diff
@@ -228,53 +231,93 @@ def validate(self, mode: str = "batch") -> List[SqlError]:
...
+
+        tasks: Set = set()
+        query_task_ids: List = []
+        errors = []
+        logger.debug(
+            "we are going to schedule %s tasks at concurrency of %s",
+            len(task_calls),
+            max_concurrency,
+        )
+        while task_calls:
```
joshtemple: When I run this with `max_concurrency = 0`, I get stuck in this loop forever and queries are never launched. `query_task_ids` is empty as defined above, so it continually waits for non-existent query task IDs to complete. Any idea what's going on here?

I wonder if you need to change `if len(query_task_ids) >= max_concurrency:` to a `>`?

DylanBaker:
One thing I realised here is that I think we lose the 250 limit on fetching queries? I can test this later on a big enough explore, but I think we will get errors when fetching the queries if it does over 293 or whatever that number is?

joshtemple:
@iserko, inspired by your PR, I did a bunch of digging on `asyncio`  and found some ways to improve this approach.

A few major changes:
- I've added a `asyncio.BoundedSemaphore` object to the validator which is used to block when the maximum number of queries are running against the data warehouse (releasing a slot once we get the result back)
- I've used an `asyncio.Queue` object to handle tracking of the running queries. This allows us to periodically check for query results as a coroutine as well, pulling all Looker query task IDs, extracting results, and pushing the running or pending queries back onto the Queue.
- I renamed `--max-concurrency` to `--query-slots`, which I think is a more intuitive name for the flag. I set the default to 10. Interested to hear your and @DylanBaker's thoughts on this though, the name is open for discussion for sure.

Let me know what you think of these changes.

PR #116, opened 2019-11-12
Run black as part of the CircleCI process

### LINKED ISSUE #113: Suggest using black formatter on the code
[Black](https://black.readthedocs.io/en/stable/) is an opinionated code formatter for Python projects
[It's use was accepted by the Django framework 5 months ago](https://github.com/django/deps/blob/master/accepted/0008-black.rst) ... I like their foreword :)

The gist of it is ... you let black deal with any formatting questions and no arguments about which comma or new line should go where or when.

As for why you would want this ... when you have external contributors with various different styles, this makes the conversation more on point about what the code is supposed to do and not how it looks.

iserko:
Fixes #113

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,12 @@ jobs:
             flake8 spectacles tests
 
       - run:
+          name: black
+          command: |
+            . venv/bin/activate
+            black --check spectacles tests
+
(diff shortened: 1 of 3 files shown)


iserko:
ready to merge @DylanBaker

PR #117, opened 2019-11-12
Handle more error variations

agile:
### Handle error w/o line number

We ran into an error with one explore which yielded a `sql_error_loc`, but did not contain a line reference:

```python

"errors": [
    {
        "message": "The PostgreSQL database encountered an error while running this query.",
        "message_details": "ERROR: column <scrubbed> does not exist\n  Position: 3178",
        "params": 'SELECT <scrubbed>',
        "edit_url": None,
        "error_pos": None,
        "level": "error",
        "sql_error_loc": {"character": 3178},
    }
],
```
### Handle errors without message_details

We encountered an error in one of our explores which only yielded a message:

```python
"errors": [
    {
        "message": 'Access filter field "scrubbed" on model "scrubbed" is unset for the current user.',
        "message_details": None,
        "params": None,
        "edit_url": None,
        "error_pos": None,
        "fatal": True,
        "level": "fatal",
        "login_required_oauth_application_id": None,
    }
],
```

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -277,12 +277,9 @@ class SqlValidator(Validator):
         if isinstance(data, dict):
             errors = data.get("errors") or [data.get("error")]
             first_error = errors[0]
-            message = first_error.get("message_details") or first_error["message"]
-            if not isinstance(message, str):
-                raise TypeError(
-                    "Unexpected message type. Expected a str, "
-                    f"received type {type(message)}: {message}"
-                )
+            message = " ".join(
+                [first_error.get("message", ""), first_error.get("message_details", "")]
+            ).strip()
             sql = data["sql"]
             error_loc = first_error.get("sql_error_loc")
             if error_loc:
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -277,7 +277,7 @@ class SqlValidator(Validator):
         if isinstance(data, dict):
             errors = data.get("errors") or [data.get("error")]
             first_error = errors[0]
-            message = first_error["message_details"]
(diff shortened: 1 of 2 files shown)


agile:
> Have you encountered any errors with `message_details` but no `message`? I notice that's what you're testing for in `error_loc_wo_line`, but in your description you only mentioned `message` with no `message_details`. Mind updating your test mock responses to match the examples you provided?

Copy pasta ;)

> More broadly, my preferred approach here would be to concatenate `message` and `message_details` into one long string and have that be what's returned as the error message. That's essentially what the Looker UI does, you get something like "The Google BigQuery database encountered an error." + "The column order_id does not exist in orders". If you can make that change, I'd be happy to merge. I'm good with your way of handling the absence of `line_number`, which is basically what was just merged in #102.

Yep, makes sense to me. The updated change should handle whichever, or both being present and combine them when they are both available.

spectacles/validators.py
```diff
@@ -277,15 +277,16 @@ def _extract_error_details(query_result: dict) -> dict:
...
-            if not isinstance(message, str):
-                raise TypeError(
-                    "Unexpected message type. Expected a str, "
-                    f"received type {type(message)}: {message}"
+            message = " ".join(
+                filter(
+                    None,
+                    (first_error.get("message"), first_error.get("message_details")),
                 )
+            )
```
joshtemple: IMO, you could do this slightly more cleanly by using the second parameter of `get`, the default response if the key is not found.

```suggestion
message = " ".join(
    [first_error.get("message", ""), first_error.get("message_details", "")]
)
```

PR #118, opened 2019-11-12
Be less restrictive with Looker version check

agile:
If the version detected is less than the defined minimum, issue a warning noting the version being used is not supported rather than refusing to run at all. (maybe it should also warn that it may crash Looker?)

We are running an older version of Looker (6.16.x) and I wanted to see if it simply would not work at all or if we were able to at least gain some utility. To my surprise, it seems to work ok, I have not run into the 500 errors (#73) that led to #81, yet.

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -106,39 +106,6 @@ class LookerClient:
 
         return response.json()["looker_release_version"]
 
-    def validate_looker_release_version(self, required_version: str) -> bool:
-        """Checks that the current Looker version meets a specified minimum.
-
-        Args:
-            required_version: Minimum instance version number (e.g. 6.22.12)
-
-        Returns:
-            bool: True if the current Looker version >= the required version
-
-        """
-        current_version = self.get_looker_release_version()
-        logger.info(f"Looker instance version is {current_version}")
-
-        def expand_version(version: str):
-            return [int(number) for number in version.split(".")]
-
-        current = expand_version(current_version)
-        required = expand_version(required_version)
-
-        # If version is provided in format 6.20 or 7, extend with .0(s)
-        # e.g. 6.20 would become 6.20.0, 7 would become 7.0.0
-        if len(current) < 3:
-            current.extend([0] * (3 - len(current)))
-
-        for current_num, required_num in zip(current, required):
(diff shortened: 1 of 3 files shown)


DylanBaker:
@agile Are you on self-hosted? 

The regression that needed #81 was introduced in 6.20 AFAIK, which makes sense that you are able to run on a previous version and makes a good argument for loosening this.

agile:
Yes, self hosted. Good to know that it's something introduced in 6.20, I believe we've been planning to upgrade to that fairly soon but maybe we can leap frog up to 6.22 or better instead.

joshtemple:
The API fix _should've_ been backported to 6.20, so it's possible it may even work for 6.20 instances. I'm comfortable removing this restriction altogether (including the warning proposed here) until we hear otherwise from people on 6.20-6.22.

@DylanBaker, thoughts?

DylanBaker:
@joshtemple Yeah, I think I am okay with that. Is there value in going back and seeing what version actually introduced the endpoints were are hitting? Are we at all worried that people are using really old versions of Looker? 

Ultimately, I think I'm okay removing for the moment, but it would be good to ascertain whether they have actually backported it. I'd also be keen to find a way with Looker to test this on an ongoing basis against a range of instance versions.

DylanBaker:
@agile Hey. As @joshtemple suggests, I think the best tact here is to just remove the exception and not have a warning. The fix should be backported so we probably don't need either. Would you mind updating the PR?

DylanBaker:
@agile I should have also mentioned, we have a Slack community for people who are using and contributing to Spectacles. If you're interested in joining, you can do so [here](https://join.slack.com/t/spectacles-ci/shared_invite/enQtODI4NTQ0NzExMTEwLTVlNzNkYzUxNDY4NmRkOWU3ZTgyMjQ1Y2ZmOTI1YTUxNGMyNDFlMDkzMzVlMjQxOTgwN2Q5ODE5NTg3NDVkYTg).

agile:
Ok, version checking removed ;) Will join slack later

joshtemple:
Thanks @agile, we really appreciate the help with this PR!

PR #121, opened 2019-11-12
Allow model names to contain periods

agile:
We are probably outliers here, but our models often have periods in their names which is not handled very well by the model/explore filtering. 

Alternatively, I wonder if we could we use another delimiter than period, such as a slash '/' instead?

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -108,7 +108,7 @@ class SqlValidator(Validator):
         selection: DefaultDict = defaultdict(set)
         for selector in selectors:
             try:
-                model, explore = selector.split(".")
+                model, explore = selector.rsplit(".", 1)
             except ValueError:
                 raise SpectaclesException(
                     f"Explore selector '{selector}' is not valid.\n"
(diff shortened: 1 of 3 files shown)


DylanBaker:
@agile I somehow didn't realise you could have periods in a model name. Do you know if it is possible to have them in explore names?

I am inclined to think that you're correct about this being an outlier instance, but we should solve it nonetheless. I do think it would be easiest to take the alternative approach and just delimit on something that can't be in model or explore names. @joshtemple Do you have a strong preference?

iserko:
Why not just separate it into two CLI parameters.
Instead of
```
--explores mymodel.myexplore
# replaced by
--model mymodel --explore myexplore
```

```
--explores mymodel.*
# replaced by
--model mymodel
```

```
--explore myexplore
# should fail if you don't specify --model as well
```

DylanBaker:
How would you want it to work if you were interested in testing explores across multiple models. 

i.e. in the current setup:

```
--explores model_one.explore_one model_two.explore_two
```

DylanBaker:
@agile It looks like a slash `/` should work. It isn't allowed in either a model or an explore name, which solved this problem. That would be our preferred solution I think.

PR #129, opened 2019-11-15
Add always false filter to queries to improve speed (particularly on Snowflake)

DylanBaker:
We run `limit 0` queries in order to test the validity of the SQL Looker generates. On lots of warehouses, that processes no data and is very fast (BigQuery and Redshift, notably).

On Snowflake, it still processes a lot of data and there isn't an `EXPLAIN` functionality we can use right now. The proposed solution is to add a `1=2` filter to all queries, which does stop it from processing data.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*  @spectacles-ci/spectacles-maintainers
+*  @DylanBaker @joshtemple
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -334,7 +334,13 @@ class LookerClient:
             explore,
             "*" if len(dimensions) > 1 else dimensions[0],
         )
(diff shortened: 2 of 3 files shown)


JarredLHumphrey:
Just an FYI that you can see the query planner output via the Snowflake instances UI at:
https://<snowflakehost>.com/console#/monitoring/queries

Once on that page, click the query id you want the plan for, and then click the `Profile` tab.

I took the liberty to run two queries to show the difference in output for `LIMIT 0` vs `WHERE 1=2`.

Query: `SELECT * FROM INFORMATION_SCHEMA.COLUMNS LIMIT 0;`
Plan:
Screen Shot 2019-11-15 at 1 59 34 PM


Query: `SELECT * FROM INFORMATION_SCHEMA.COLUMNS WHERE 1=2 LIMIT 0;`
Plan: 
Screen Shot 2019-11-15 at 1 59 52 PM

Issue #130, opened 2019-11-15
Print the total run time at completion of validation


joshtemple:
Spectacles should time itself and print out the total runtime at the end when running a validator.

An example of this would be:
```
Completed SQL validation in 38 seconds
```

As a bonus, elapsed time should be printed in the most human-readable form, e.g. 2 minutes and 47 seconds instead of 167 seconds.

PR #133, opened 2019-11-16
Issue #130: Print the total run time at completion of validation

Hawk94:
Adds a `@log_time` decorator that measures the time taken to run a given function and logs the result. Currently decorates the SqlValidator validate method to log the time taken to run the SQL validation.

Screenshot 2019-11-16 at 19 16 08


DylanBaker:
@Hawk94 We can close this one, correct?

Hawk94:
correct

PR #134, opened 2019-11-17
Feature/log time assert

Hawk94:
Extend log_time decorator to handle logging from multiple different validators. Also align output from `spectacles assert` & `spectacles sql` commands:

Screenshot 2019-11-17 at 18 55 16

Screenshot 2019-11-17 at 18 55 10


DylanBaker:
@Hawk94 Thanks for this! It looks good. 

One thing I can't decide is if the time should be how long the whole process takes, or just the validate step. There's a bit of time spent building up the LookML project, that doesn't get included if we just wrap the `validate` function. @joshtemple Do you have an opinion on what it should be measuring?

PR #139, opened 2019-11-25
Cancel running queries when Spectacles gets interrupted

### LINKED ISSUE #88: Kill running queries on error or KeyboardInterrupt
If a `KeyboardInterrupt` occurs or an error happens while we're running query tasks in the SQL validator, we should kill any running query tasks [via the API](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/query#kill_running_query).

DylanBaker:
Closes #88

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -130,6 +130,11 @@ def handle_exceptions(function: Callable) -> Callable:
                 + "\n"
             )
             sys.exit(error.exit_code)
+        except KeyboardInterrupt as error:
+            logger.debug(error, exc_info=True)
+            logger.info("ctrl-c")
+            logger.info("Encountered KeyboardInterrupt.")
+            sys.exit(1)
         except Exception as error:
             logger.debug(error, exc_info=True)
             logger.error(
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -132,8 +132,7 @@ def handle_exceptions(function: Callable) -> Callable:
             sys.exit(error.exit_code)
         except KeyboardInterrupt as error:
             logger.debug(error, exc_info=True)
-            logger.info("ctrl-c")
-            logger.info("Encountered KeyboardInterrupt.")
+            logger.info("Spectacles was manually interrupted.")
             sys.exit(1)
         except Exception as error:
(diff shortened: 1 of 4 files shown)


DylanBaker:
@joshtemple This is a rough initial implementation. As discussed, as far as I can tell, it does cancel the queries as expected, except for when they are removed from the `running_query_tasks` queue in `_get_query_results`.

Maybe that's good enough for a first pass, but I'd obviously much rather get this done properly. It's no longer such as issue now that we have the `--concurrency` flag.

Very open to suggestions on ways to tackle this. I'll obviously also  give it further thought.

spectacles/validators.py
```diff
@@ -244,13 +261,33 @@ def validate(self, mode: str = "batch") -> List[SqlError]:
...
+                    self.client.cancel_query_task(session, query_task_id)
+                )
+                cancel_query_tasks.append(task)
 
-        await session.close()
+            await asyncio.gather(*cancel_query_tasks)
 
-        return errors
+            raise SpectaclesException(
+                "Spectaces was interrupted. "
```
joshtemple: Can we prepend 1-2 newlines here so it doesn't look like this?
```
^CSpectaces was interrupted.```

File: spectacles/cli.py [hunk 2]
```diff
@@ -130,6 +130,11 @@ def wrapper(*args, **kwargs):
                 + "\n"
             )
             sys.exit(error.exit_code)
+        except KeyboardInterrupt as error:
+            logger.debug(error, exc_info=True)
+            logger.info("ctrl-c")
```
joshtemple: Don't think we need this, plus it's not always Ctrl+C.

spectacles/cli.py
```diff
@@ -130,6 +130,11 @@ def wrapper(*args, **kwargs):
...
+            logger.info("Encountered KeyboardInterrupt.")
```
joshtemple: Something a bit more readable like, "Spectacles was manually interrupted."

spectacles/validators.py
joshtemple: Also, to match above, can we do "Spectacles was manually interrupted."?

PR #144, opened 2019-11-27
Handle access filter error

agile:
:man_facepalming: 

This was proposed to be fixed via #117, but an incomplete example in that PR led to an incomplete solution. This time I updated the example to use a snippet taken directly from the error that was logged.

Note that there is also no `sql` key in this error which is now handled as well.

### DIFF
--- a/tests/test_sql_validator.py
+++ b/tests/test_sql_validator.py
@@ -141,32 +141,15 @@ def test_extract_error_details_error_non_str_message_details(validator, project)
         validator._extract_error_details(query_result)
 
 
-def test_extract_error_details_for_unset_access_filter(validator, project):
-    message = (
-        'Access filter field "foo.bar" on model "baz.qux" '
-        + "is unset for the current user."
-    )
-    sql = None
+def test_extract_error_details_no_message_details(validator, project):
+    message = "An error message."
     query_result = {
         "status": "error",
-        "data": {
-            "errors": [
-                {
-                    "message": message,
-                    "message_details": None,
-                    "params": None,
-                    "edit_url": None,
-                    "error_pos": None,
-                    "fatal": True,
-                    "level": "fatal",
-                    "login_required_oauth_application_id": None,
-                }
-            ]
-        },
+        "data": {"errors": [{"message": message, "message_details": None}]},
     }
     extracted = validator._extract_error_details(query_result)
     assert extracted["message"] == message
-    assert extracted["sql"] == sql
+    assert extracted["sql"] is None
 
 
 def test_extract_error_details_error_loc_wo_line(validator, project):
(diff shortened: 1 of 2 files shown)


DylanBaker:
@agile This looks good to me. The test just needs a bit of re-formatting to be aligned with Black.

agile:
Weird, I use black but must have inadvertently updated to a newer version at some point. 
```
$ black --version && black --check spectacles tests
black, version 19.10b0
All done! ✨ 🍰 ✨
17 files would be left unchanged.
```
Downgraded back down and fixed it. Thanks!

tests/test_sql_validator.py
```diff
@@ -141,12 +141,28 @@ def test_extract_error_details_error_non_str_message_details(validator, project)
         validator._extract_error_details(query_result)
 
 
-def test_extract_error_details_error_loc_wo_msg_details(validator, project):
-    message = "An error message."
-    sql = "SELECT * FROM orders"
+def test_extract_error_details_for_unset_access_filter(validator, project):
```
joshtemple: Can we make this more of a generalized test case title? That is, what's the combination of fields that's being tested here? `message` is present but `message_details` is not? Is it that `sql` is not present? Want to give the test a title that describes what's being tested in an abstract sense rather than the exact error that you received.

joshtemple: Also, I believe we can remove the fixture `project` here since I don't see it being used.

tests/test_sql_validator.py
```diff
@@ -141,12 +141,28 @@ def test_extract_error_details_error_non_str_message_details(validator, project)
...
+    sql = None
     query_result = {
         "status": "error",
-        "data": {"errors": [{"message": message}], "sql": sql},
+        "data": {
+            "errors": [
+                {
+                    "message": message,
+                    "message_details": None,
+                    "params": None,
```
joshtemple: Can you remove these other response keys (`params`, `edit_url`, etc.) unless they're being used in the test? Goal is to make this an abstract representation of exactly what's being tested, not the verbatim response that generated the problem being fixed here!

tests/test_sql_validator.py
agile: Updated, let me know if there's anything else

tests/test_sql_validator.py
```diff
@@ -141,19 +141,19 @@ def test_extract_error_details_error_non_str_message_details(validator, project)
         validator._extract_error_details(query_result)
 
 
-def test_extract_error_details_error_loc_wo_msg_details(validator, project):
+def test_extract_error_details_no_message_details(validator):
     message = "An error message."
-    sql = "SELECT * FROM orders"
+    sql = None
```
joshtemple: Can remove this line and do `assert extracted["sql"] is None` at the bottom.

PR #146, opened 2019-12-02
Allow users to specify logs location

### LINKED ISSUE #141: Provide flag to specify where logs should be written to
Logs are currently written to the `logs/` directory. We should provide the ability to specify where the logs get written.

DylanBaker:
Closes #141

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -161,7 +161,7 @@ def set_file_handler(directory: str) -> None:
     LOG_FILEPATH = Path(log_directory / LOG_FILENAME)
     log_directory.mkdir(exist_ok=True)
 
-    fh = logging.FileHandler(log_filepath)
+    fh = logging.FileHandler(LOG_FILEPATH)
     fh.setLevel(logging.DEBUG)
 
     formatter = FileFormatter("%(asctime)s %(levelname)s | %(message)s")
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -11,7 +11,6 @@ from spectacles.client import LookerClient
 from spectacles.exceptions import SpectaclesException, ValidationError
 from spectacles.logger import GLOBAL_LOGGER as logger, FileFormatter
 import spectacles.printer as printer
-from pathlib import Path
 
 LOG_FILENAME = "spectacles.log"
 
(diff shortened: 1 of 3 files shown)


DylanBaker:
@joshtemple Think there's probably something basic I'm missing here (possibly). 

mypy doesn't like that `LOG_FILEPATH` isn't declared globally before I call it as a global variable in `set_file_handler`. As a result, when I use it there, it gives the following error:

```spectacles/cli.py:161: error: Name 'LOG_FILEPATH' is not defined
spectacles/cli.py:164: error: Name 'LOG_FILEPATH' is not defined
```

If I set something like `LOG_FILEPATH = Path("")` at the top of the file, the error goes away, but seems like the wrong way to fix it.

spectacles/cli.py
```diff
@@ -150,14 +153,32 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def set_file_handler(directory: str) -> None:
+
+    global LOG_FILEPATH
```
joshtemple: Why does this need to be global? Only see it being used in this function.

spectacles/cli.py
```diff
@@ -9,8 +9,11 @@
 from spectacles.runner import Runner
 from spectacles.client import LookerClient
 from spectacles.exceptions import SpectaclesException, ValidationError
-from spectacles.logger import GLOBAL_LOGGER as logger, LOG_FILEPATH
+from spectacles.logger import GLOBAL_LOGGER as logger, FileFormatter
 import spectacles.printer as printer
+from pathlib import Path
```
joshtemple: Can you move this to the top? Stdlib imports should go before local imports.

spectacles/cli.py
```diff
@@ -273,6 +294,13 @@ def _build_base_subparser() -> argparse.ArgumentParser:
         help="Display debug logging during spectacles execution. \
             Useful for debugging and making bug reports.",
     )
+    base_subparser.add_argument(
+        "--log-directory",
+        action=EnvVarAction,
+        env_var="LOG_DIRECTORY",
```
joshtemple: Can we do `SPECTACLES_LOG_DIR` to avoid any conflict with non-Spectacles environment variables? Best practice to be app-specific with these.

spectacles/cli.py
```diff
@@ -273,6 +294,13 @@ def _build_base_subparser() -> argparse.ArgumentParser:
         help="Display debug logging during spectacles execution. \
             Useful for debugging and making bug reports.",
     )
+    base_subparser.add_argument(
+        "--log-directory",
```
joshtemple: `--log-dir` work for this? Think it's a common enough abbreviation I'd be okay with it for convenience.

spectacles/cli.py
DylanBaker: It's used in the exception handler to direct people to logs location.

Issue #96, opened 2019-11-05
Change client methods to have one to one relationship with endpoints


DylanBaker:
Methods like `update_session` currently do multiple things. We should split these out and move the logic on running them together to the runner.

PR #156, opened 2020-01-30
Allow specifying ref to checkout

agile:
These changes add a `ref` option which may also be set via `LOOKER_GIT_REF` environment variable. When set, ref will be passed to Looker's `git_branch` call to explicitly request the referenced commit be checked out.

~~This resolves an annoyance I've been experiencing while cleaning up some looks where I run `spectacles` which identifies several dimensions with issues. I fix, commit and push the changes but rerunning `spectacles` continues to report them until I go into looker and manually pull the changes into the space. I'd hoped the `--remote-reset` option would resolve this, but it does not. I left that option unchanged as it may be useful in other scenarios.~~  

The `--remote-reset` option does fix this scenario, actually. This `ref` feature is still useful since it can be used to by CI to ensure the state of the repo being tested matches the commit that triggered the build.

While I was in there, I also segregated the calls to be closer to 1:1 with the API. (#96)

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -203,6 +203,7 @@ def main():
             args.mode,
             args.remote_reset,
             args.concurrency,
+            args.ref,
         )
     elif args.command == "assert":
         run_assert(
@@ -214,6 +215,7 @@ def main():
             args.port,
             args.api_version,
             args.remote_reset,
+            args.ref,
         )
 
 
@@ -364,6 +366,14 @@ def _build_sql_subparser(
         help="The branch of your project that spectacles will use to run queries.",
     )
     subparser.add_argument(
+        "--ref",
+        action=EnvVarAction,
+        env_var="LOOKER_GIT_REF",
+        help="A specific Git ref (sha) to reset the branch in Looker to. \
+            WARNING: This will delete any uncommited changes in the user's \
+            workspace.",
+    )
+    subparser.add_argument(
         "--explores",
         nargs="+",
         default=["*/*"],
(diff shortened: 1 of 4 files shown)


DylanBaker:
Hey @agile. Thanks for this. 

As you've said, I'm surprised that `--remote-reset` doesn't solve this problem. When you do run this with `--remote-reset` and find that the errors have persisted, what does the Looker dev UI show when using the Spectacles user?

agile:
> As you've said, I'm surprised that `--remote-reset` doesn't solve this problem. When you do run this with `--remote-reset` and find that the errors have persisted, what does the Looker dev UI show when using the Spectacles user?

Hmm, I think the `--remote-reset` was actually effective and I was just confused looking at my own workspace vs the one spectacles was using while working with some tricky columns being used in several explores. :man_facepalming: Updated the description to strike that statement.

DylanBaker:
Great. I would have been worried if that wasn't working, as it's exactly what the flag is for. I agree though that it's useful to be able to specify the git ref anyway. Will review shortly. Thanks again.

joshtemple:
Closing this because it has some unintended consequences due to the way the Looker endpoint [Update Project Git Branch](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/project#update_project_git_branch) works. Specifying `ref` in the request body not only hard resets the Looker local copy of the branch to the specified commit hash, it also force pushes the result to remote, which permanently deletes anything after the specified commit from both repositories.

This is the line in the API docs that indicates the danger of providing `ref`:
>**DANGER** hard reset will be force pushed to the remote. Unsaved changes and commits may be permanently lost.

A safer way to achieve this same result might be to create a temporary branch for every Spectacles run to freeze the branch state to when the run was launched or queued. We're currently looking into that approach.

Thanks for the contribution regardless @agile! We're definitely interested in making this concept work, so stay tuned for an alternate approach.

PR #168, opened 2020-03-22
Setting dependent project branches to master

### LINKED ISSUE #151: Control branches on dependent projects
Some Spectacles users reference other LookML projects as part of their Looker setup. They do so by declaring the project in a [manifest file](https://docs.looker.com/reference/manifest-reference).

Currently, if Spectacles is used on a project that leverages a second project, it will use a version of that second project on whatever branch that user was last using when developing that project. This can cause false positives and false negatives when testing. 

Ideally, Spectacles would use the master version of the dependencies declared in the manifest file.

We can't _actually_ use master because the master branch is not available while in dev mode. One option is to create a new branch from master and to set the user to that branch. At the end of the run, we delete the branch.

Obvious downside: if the run doesn't complete fully, the branch will be left there (though ideally would be easily findable with a name like 'spectacles_tmp_somehash`.

DylanBaker:
closes #151


DylanBaker:
@joshtemple Would be great to get your eyes on this. 

There's a few things outstanding:
* This work uses the `client.update_session` method, which is actually overkill because we only need it to move the branch, not do the other things that method does. I think I could use this opportunity to split that method up to have a one-to-one relationship with Looker API endpoints?
* The 'Checked Out'  logger message at the end of `client.update_session` currently is set to `info`. As a result, the branch that we are moving the other projects to in the cleanup gets called out. Is the right thing to move to debug or control dynamically depending on whether it's the setup or cleanup?
* Need to add tests

DylanBaker:
@joshtemple I'm clearly doing something not-quite-right(tm) in how I'm setting up the decorator, based on the mypy error, but not _quite_ sure what the solution it.

joshtemple:
After some rebasing, do you mind closing this and re-opening against `master`? Think it will be cleaner.

PR #172, opened 2020-03-29
Exclude explores

DylanBaker:


DylanBaker:
@joshtemple I think I made a bit of a meal of rebasing that, somehow, but this should now be good.

spectacles/cli.py
```diff
@@ -371,6 +372,15 @@ def _build_sql_subparser(
             The '*' wildcard selects all models or explores. For instance,\
             'model_name/*' would select all explores in the 'model_name' model.",
     )
+    subparser.add_argument(
+        "--exclude",
+        nargs="+",
+        default=[],
+        help="Specify the explores spectacles should exclude when testing. \
+            List of selector strings in 'model_name/explore_name' format. \
+            The '*' wildcard selects all models or explores. For instance,\
```
joshtemple:
```suggestion
            The '*' wildcard excludes all models or explores. For instance,\
```

spectacles/validators.py
```diff
@@ -194,9 +195,25 @@ def build_project(self, selectors: List[str]) -> None:
...
-        for model in selected_models:
+        excluded_explores = {}
+        for model in excluded_models:
             # Expand wildcard operator to include all specified or discovered explores
+            excluded_explore_names = exclusion[model.name]
+            if "*" in excluded_explore_names:
+                excluded_explore_names.remove("*")
+                excluded_explore_names.update(
+                    set(explore.name for explore in model.explores)
+                )
```
joshtemple: If there's a situation with `model_A.explore_A` and `model_B.explore_A`, and the `--exclude` argument is `model_A.explore_A`, would this still preserve `explore_A` in `model_B`? Any way we can add this case to the unit tests?

DylanBaker: It will still preserve it. The explores to exclude are stored in lists against a key of the model name. We then only check against the explores within a given model: https://github.com/spectacles-ci/spectacles/pull/172/files#diff-cbee2ab0268f8f001868af4ca066ba04R228-R232

I'll definitely add some tests to confirm that though.

DylanBaker: @joshtemple I've confirmed this is the case and have added an additional test `test_build_project_one_ambiguous_explore_excluded` to verify so.

I also actually re-factored some of the other tests to make the selection easier.

PR #174, opened 2020-03-31
Return objects from runner methods, not dictionaries

### LINKED ISSUE #142: Have validators return the full LookML project, not just errors.

joshtemple:
Closes #142.

Changes the Runner validation methods to return native objects (e.g. `spectacles.lookml.Project`) instead of lists of dictionaries.

@DylanBaker, the `validate_data_tests` method only returns a list of `DataTestError` currently because assertions are not directly associated with the LookML structure. I'm curious to know if we should still try and slot these into a Model instead of just returning a list of errors. It may be helpful to have the consistency of always returning a Project object.

This would allow us to break out the `build_project` call and run that at `Runner` instantiation instead of individually for each validator.

FYI, I'm going to leave Project serialization for a separate PR.

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -484,7 +484,7 @@ def run_sql(
         api_version,
         remote_reset,
     )
-    errors = runner.validate_sql(explores, mode, concurrency)
+    project = runner.validate_sql(explores, mode, concurrency)
     if errors:
         for error in sorted(errors, key=lambda x: x["path"]):
             printer.print_sql_error(error)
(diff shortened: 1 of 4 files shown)


DylanBaker:
@joshtemple Yeah, I think ideally we should always ideally return the Project. By having them all use the `build_project` method, we could also then apply the `--explores` and `--exclude` flags to the assert command.

DylanBaker:
@joshtemple I'd be happy to leave that to a subsequent PR though?

PR #175, opened 2020-04-01
Adding anonymised Segment tracking

DylanBaker:
Wanted to create a draft of what tracking might look like. 

I settled on Segment because:
* GA and Heap don't seem to work well with exclusively server-side events
* Snowplow doesn't have a hosted free tier 
* Segment has pretty good free tier for startups with minimal funding

Ultimately, I think I'd like to use Snowplow, but this should be an easy way to get us going. 

I wanted to get eyes on a possible implementation before adding testing/etc. 

@joshtemple @Hawk94 You should have both received an invite to Segment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ requests==2.22.0
 colorama==0.4.1
 PyYAML==5.1.2
 backoff==1.8.0
+analytics-python==1.2.9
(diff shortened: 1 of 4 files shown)


DylanBaker:
All the tracking here is anonymised. We send:
* A hash of the base URL
* A hash of the project
* The command that is run
* A unique 'invocation_id' that is random and generated by Spectacles

As such, we won't be able to see anything about who is running Spectacles. We will just be able to see that Spectacles was run against an anonymised url and project.

DylanBaker:
We will also add do not track functionality so that users can opt-out.

DylanBaker:
@joshtemple I think this is good for review.

PR #179, opened 2020-04-04
Log failing SQL to file

### LINKED ISSUE #84: Failing SQL is not saved to file
After some architectural changes, failing SQL queries are no longer being logged to file. We need to reintroduce this feature. It should probably live in `cli.py`, using the value of the `sql` key in the dictionary returned by the runner.

Previously, in single-dimension mode, spectacles would create a folder per explore with a `.sql` file for each dimension query. In batch mode, it would create a `.sql` file per explore in the base directory.

These files were all created in the `logs` directory.

DylanBaker:
closes #84 

This PR grabs the `sql` variable on the `SqlError` and logs the query to the logs and to a file in the logs directory.

Initially, I wrote it that the `log_failing_sql` function didn't take a directory, and just used the `LOG_DIRECTORY` global variable. I was having some difficulty getting the tests to work though as I couldn't seem to overwrite that variable and log the queries to a temp directory. Very open to feedback/advise on how to improve that.


tests/test_cli.py
```diff
@@ -181,3 +183,46 @@ def test_bad_config_file_parameter(mock_parse_config, clean_env, parser):
 def test_parse_remote_reset_with_assert(env, parser):
     args = parser.parse_args(["assert", "--remote-reset"])
     assert args.remote_reset
+
+
+def test_logging_failing_explore_sql(tmpdir):
```
joshtemple, line 188: This fixture is cool, I didn't know about this!

tests/test_cli.py
```diff
@@ -181,3 +183,46 @@ def test_bad_config_file_parameter(mock_parse_config, clean_env, parser):
...
+        message="example error message",
+        sql="select example_explore.example_dimension_1 from model",
+    )
+
+    log_failing_sql(error, tmpdir, "explore_model", "example_explore")
+    query_directory = Path(tmpdir / "queries")
+    query_file = Path(query_directory / "explore_model__example_explore.sql")
+    content = open(query_file).read()
+
+    assert os.path.exists(query_directory)
```
joshtemple, line 200: Let's drop the `os` import and use `Path.exists()` since you're already using pathlib.

spectacles/cli.py
```diff
@@ -440,6 +442,32 @@ def _build_assert_subparser(
     )
 
 
+def log_failing_sql(
```
joshtemple, line 445: Maybe this is something we do in a future PR, but it would be _really_ cool to put the error message and LookML link (just like we do in the CLI output) in a comment block at the top of this file.

If it would be useful, we could also add a timestamp for when the query was run.

For example:
```sql
/*
The Google BigQuery Standard SQL database encountered an error while running
this query. Failed to retrieve data - Name numbers_of_orders not found inside
dim_customers at [3:23]

LookML: https://spectacles.looker.com/projects/jaffle_shop/files/views%2Fdim_customers.view.lkml?line=46

Query executed at 2020-04-09 22:24:19T00:00
*/

SELECT
  dim_customers.numbers_of_orders  AS dim_customers_number_of_orders
FROM analytics.fct_orders  AS fct_orders
LEFT JOIN analytics.dim_customers  AS dim_customers ON fct_orders.customer_id = dim_customers.customer_id

WHERE 1 = 2
GROUP BY 1
ORDER BY 1
LIMIT 0
```

spectacles/cli.py
```diff
@@ -157,10 +158,11 @@ def wrapper(*args, **kwargs):
 def set_file_handler(directory: str) -> None:
 
     global LOG_FILEPATH
+    global LOG_DIRECTORY
```
joshtemple, line 161: Don't think we need both of these to be global. Can remove `LOG_FILEPATH` as global and just use `LOG_DIRECTORY / LOG_FILENAME` as a replacement.

spectacles/cli.py
```diff
@@ -157,10 +158,11 @@ def wrapper(*args, **kwargs):
...
 
-    log_directory = Path(directory)
-    LOG_FILEPATH = Path(log_directory / LOG_FILENAME)
-    log_directory.mkdir(exist_ok=True)
+    LOG_DIRECTORY = Path(directory)
+    LOG_FILEPATH = Path(LOG_DIRECTORY / LOG_FILENAME)
```
joshtemple, line 164: Don't need to wrap this in Path anymore, can just do `LOG_DIRECTORY / LOG_FILENAME` since the directory is already a Path object.

spectacles/cli.py
```diff
@@ -5,16 +5,17 @@
...
 from spectacles.runner import Runner
 from spectacles.client import LookerClient
-from spectacles.exceptions import SpectaclesException, ValidationError
+from spectacles.exceptions import SpectaclesException, ValidationError, SqlError
 from spectacles.logger import GLOBAL_LOGGER as logger, FileFormatter
 import spectacles.printer as printer
 
 LOG_FILENAME = "spectacles.log"
 LOG_FILEPATH = Path()
+LOG_DIRECTORY = Path("logs")
```
joshtemple, line 18: I don't think we actually need this since it's already the default value for the `log_dir` CLI arg, which is passed to `set_file_handler`.

DylanBaker: The `log_failing_sql` function is going to require knowledge of either the directory it should create the 'queries' directory in, or the directory that it should dump the files themselves.

I agree having `LOG_DIRECTORY` is clunky. My suggestion is to actually declare something like `QUERY_FILEPATH` as a global variables that also gets set in `set_file_handler`. We then just pass that `QUERY_FILEPATH` to the `log_failing_sql` function. Would you prefer that approach?

spectacles/cli.py
DylanBaker: I love this. I'm going to suggest I make an issue and add it in a subsequent PR?

spectacles/cli.py
DylanBaker: (Current code reflects this new approach if you want to have a look.)

spectacles/cli.py
joshtemple: Fine by me!

DylanBaker: https://github.com/spectacles-ci/spectacles/issues/183

blahblah/queries/jaffle_shop__fct_orders.sql
```diff
@@ -0,0 +1,37 @@
+SELECT
```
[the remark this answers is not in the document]
DylanBaker: Ooof. Yeah. I'll remove...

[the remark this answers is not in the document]
DylanBaker: Oh. Actually realise I've already removed.

PR #180, opened 2020-04-04
Add Query URL to SqlErrors

DylanBaker:
This PR adds the 'share_url' to every query and every SqlError. This is needed for the web UI.

We might want to additionally share the URL in the CLI, but this PR does not achieve that.


spectacles/client.py
```diff
@@ -349,7 +349,7 @@ def create_query(self, model: str, explore: str, dimensions: List[str]) -> int:
             "*" if len(dimensions) != 1 else dimensions[0],
             query_id,
         )
-        return query_id
+        return result
```
joshtemple, line 352: Confused how this got past mypy, but you'll need to change the return type hint of this message to match.

spectacles/exceptions.py
```diff
@@ -16,13 +16,15 @@ def __init__(
         path: str,
         message: str,
         sql: str,
+        query_url: str,
```
joshtemple, line 19: Are you okay with calling this `explore_url` for clarity?

spectacles/validators.py
```diff
@@ -311,19 +313,17 @@ def _create_queries(self, mode: str) -> List[Query]:
     def _create_explore_query(self, explore: Explore, model_name: str) -> Query:
         """Creates a single query with all dimensions of an explore"""
         dimensions = [dimension.name for dimension in explore.dimensions]
-        query_id = self.client.create_query(model_name, explore.name, dimensions)
-        return Query(query_id, lookml_ref=explore)
+        query = self.client.create_query(model_name, explore.name, dimensions)
+        return Query(query["id"], lookml_ref=explore, query_url=query["share_url"])
```
joshtemple, line 317: Since we're returning the full response from the API now, could you add an `@classmethod` alternate constructor to the Query class called `from_json` that accepts the full response and builds a query? This is similar to the LookML objects. Nice and clean and you don't have to pass args really you can just pass that `query` variable directly like this:
```
return Query.from_json(query)
```

DylanBaker: We would still need to pass the `lookml_ref` manually, right? It's not part of the JSON response.

spectacles/exceptions.py
DylanBaker: Yeah 👍 I'll change that.

spectacles/validators.py
joshtemple: Oh you're right. I guess it's a question of is
```
return Query.from_json(query, lookml_ref=explore)
```
Really that much better. It's your call, I might be being overly picky here.

DylanBaker: For the moment, I think I'd rather leave as is. It feels like overkill if we only pass two arguments, though I think if we add any more I agree we should add the functionality.

DylanBaker:
@joshtemple I think this is good to go.

PR #186, opened 2020-04-28
Fix bug not respecting single or hybrid mode

joshtemple:
This PR introduces a few fixes:

- Moves `log_failing_sql` so it sits in `logger.py`, adds a separate test module for log testing, and prepends "Test SQL: " before the filename in the output so it's clear what the path is for.
- Fixes a bug where dimension queries were referencing their explores, not dimensions, effectively breaking single mode.
- Fixes another bug in cli.py where the dimension-level errors were not being printed in hybrid mode.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -332,7 +332,7 @@ class SqlValidator(Validator):
         for dimension in explore.dimensions:
             query = self.client.create_query(model_name, explore.name, [dimension.name])
             query = Query(
-                query["id"], lookml_ref=explore, explore_url=query["share_url"]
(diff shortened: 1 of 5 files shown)


spectacles/validators.py
```diff
@@ -332,7 +332,7 @@ def _create_dimension_queries(
         for dimension in explore.dimensions:
             query = self.client.create_query(model_name, explore.name, [dimension.name])
             query = Query(
-                query["id"], lookml_ref=explore, explore_url=query["share_url"]
+                query["id"], lookml_ref=dimension, explore_url=query["share_url"]
```
DylanBaker, line 335: Think this was my mistake. Apologies!

Issue #164, opened 2020-03-17
Getting Looker warnings as errors via the API


DylanBaker:
I am getting the following error via Spectacles:

```
Note: This query contains derived tables with conditional SQL for Development
Mode. Query results in Production Mode might be different.
```

It is occurring when I run it against a view with the following derived table: 

```
select * from
-- if dev -- {{_user_attributes['sandbox_schema']}}.fct_segment_tracks
-- if prod -- {{_user_attributes['dbt_schema']}}.fct_segment_tracks
where event_source in ('iOS','Android','Web and Server')
```

The error seems to be more of a 'warning' than an error, but the API doesn't distinguish between the two.

<img width="943" alt="Screenshot 2020-03-17 at 18 06 05" src="https://user-images.githubusercontent.com/10001014/76906283-0062df80-687a-11ea-98e3-5ac2d1de67d5.png">

My initial idea is to collate a list of these non-error errors and just exclude them, but I'm also going to check with Looker to see if there's other info we get that will be more informative.

DylanBaker:
The error returned from API for a normal SQL error: 
```            "errors": [
                {
                    "message": "The Google BigQuery Standard SQL database encountered an error while running this query.",
                    "message_details": "Failed to retrieve data - Name customer_id_typo not found inside dim_customers at [36:67]",
[...]

PR #193, opened 2020-05-02
Ignore 'warning' errors and report all errors

### LINKED ISSUE #140: Describe all errors in query result, not just the first
Spectacles only displays the first query error returned by the Looker API. Sometimes the API returns multiple results. For example, Looker might return a warning about conditional SQL, followed by the actual error from the data warehouse.

In this case, displaying only the first error is not helpful because we miss out on the actual SQL error message. We need a thoughtful way to combine multiple errors into a single message per query.

DylanBaker:
This is heavily based on #159. That PR was based on the codebase before we removed the async functionality and it was going to be easier to do again rather than re-factor. 

This PR achieves two things:
* Allows Spectacles to report multiple errors for a given dimension or explore
* Ignores the `"Note: This query contains derived tables with conditional SQL for Development Mode. Query results in Production Mode might be different."` error that is actually warning.

The most meaningful change in this PR is that instead of the LookML objects having an `error` attribute, it is now an `errors` attribute.

Closes #140, #164  

#### To do:
- [x] More testing

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -534,17 +534,21 @@ def run_sql(
         for model in iter_errors(project.models):
             for explore in iter_errors(model.explores):
                 if explore.errors and mode == "batch":
+                    file_path = log_sql_error(
+                        explore.errors, log_dir, model.name, explore.name
+                    )
                     for error in explore.errors:
                         printer.print_sql_error(error)
-                        file_path = log_sql_error(
-                            error, log_dir, model.name, explore.name
-                        )
                 else:
                     for dimension in iter_errors(explore.dimensions):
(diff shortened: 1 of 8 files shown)


DylanBaker:
@joshtemple I _think_ to test this fully, I'll need to edit the `eye-exam` repo so that it has some conditional logic and explores that generate two errors, and the re-record the cassettes. Is that right?

DylanBaker:
@joshtemple I'm going to need to review this given other changes merged in. Tried rebasing but ran into a number of problems. Will work on it tomorrow.

DylanBaker:
Going to re-work this in a new PR.

PR #194, opened 2020-05-05
Refactor to return a consistent data structure from validators

joshtemple:
The goal of this PR is to refactor the data structure returned by each validator to be more generic across validators.

First, I've made changes to the ValidationError exception and its children.
- `ValidationError` subclasses follow a general structure now, with validator-specific data stuffed into a `metadata` dictionary attribute.
- `ValidationError` now inherits from `GenericValidationError`. This was done so we have something that doesn't require arguments to raise in the CLI. Previously we were raising the last `ValidationError` subclass returned by validator via the runner. Now we don't have an actual exception object to raise (only dictionaries), so we need something generic with an exit code.

Next, I had to add some fields to the LookML objects.
- Dimension and Explore objects are passed around in the SQL Validator without knowledge of where they sit in the project hierarchy (i.e. without knowledge of their parents). This made populating the desired error data structure (with model and explore names) impossible.
- I've added `model_name` and `explore_name` to the `Dimension` class and `model_name` to the `Explore` class. `project_name` was already an attribute of `Model`.
- These fields are populated by the `from_json` methods

I updated what's returned by each validator to match the desired data structure
- For SQL validation, I added a method `get_results` to the `Project` object. @DylanBaker, I'm open to discussing if this is the right place to put this or if this is a transformation we should apply to project from _within_ the validator.
- For data test validation, I added code to the `validate` method that does this
- I updated code in the `logger`, `printer`, and `cli` modules to handle this new data structure (generally it's changing methods to accept a dictionary rather than a `ValidationError` instance

I made some quality-of-life improvements to testing
- Moved the `load` function out of `test_lookml.py` and into a `tests.utils` module
- I added a VCR test for the DataTestValidator
- I added a JSON Schema to validate the response from each validator and added tests to do so
- Centralize LookML fixtures in `confest.py`


spectacles/logger.py
```diff
@@ -57,26 +56,21 @@ def format(self, record):
...
 ) -> Path:
-
     file_name = (
-        model_name
+        model
         + "__"
-        + explore_name
-        + ("__" + dimension_name if dimension_name else "")
+        + explore
+        + ("__" + dimension.replace(".", "_") if dimension else "")
```
DylanBaker, line 65: Do we need to do this `replace()` on the model and explore? Can they have `.` in them?

spectacles/validators.py
```diff
@@ -67,28 +67,68 @@ def __init__(self, client: LookerClient, project: str):
...
+
+        tested = reduce_result(tested)
+        for test in tested:
+            printer.print_validation_result(
+                passed=test["passed"], source=f"{test['model']}.{test['explore']}"
+            )
+
+        passed = min(test["passed"] for test in tested)
+        return {
+            "validator": "sql",
```
DylanBaker, line 127: I think `data` or `assert` here, not `sql`?

spectacles/logger.py
joshtemple: Yeah, remember how we had to change the selector syntax to `model/explore` because someone apparently put dots in their names?

joshtemple: Good idea.

PR #197, opened 2020-05-10
Add functionality to test specific ref

DylanBaker:
This PR builds off of work from three others, so is not ready until those get merged in. (Those PRs are #196, #192, and #189, which would have all conflicted with this work if it had been done against master.)

This PR adds:
* a `commit_ref` argument, which specifies which commit to test.
* functionality to the runner to create a new branch if the commit ref argument is provided. We can't do it on the existing branch as we risk losing work.
* a decorator `cleanup_temp_branches` which deletes the temp branch created during the runner setup


spectacles/client.py
```diff
@@ -119,98 +119,82 @@ def get_looker_release_version(self) -> str:
...
+        response = self.session.patch(url=url, json=body, timeout=TIMEOUT_SEC)
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            raise LookerApiError(
+                name="unable-to-update-workspace",
+                title="Couldn't update the session's workspace.",
+                status=response.status_code,
+                detail=(
+                    "If you have any unsaved work on the branch "
```
DylanBaker, line 141: Add which workspace was being update to.

DylanBaker:
@joshtemple Fixed tests. This one is actually now good to go.

spectacles/cli.py
```diff
@@ -375,23 +377,24 @@ def _build_validator_subparser(
...
+    group.add_argument(
         "--remote-reset",
         action=EnvVarStoreTrueAction,
         env_var="SPECTACLES_REMOTE_RESET",
         help="When set to true, the SQL validator will tell Looker to reset the \
             user's branch to the revision of the branch that is on the remote. \
             WARNING: This will delete any uncommited changes in the user's workspace.",
     )
-    base_subparser.add_argument(
-        "--import-projects",
```
joshtemple, line 389: This argument was deleted somehow (maybe during rebasing?) and needs to be restored. Spectacles won't run with this PR at the moment.

Issue #201, opened 2020-05-13
Circle runs our test suite twice


joshtemple:
The `pytest` and `coverage` statements are duplicate which nearly doubles the length of our build. We can call probably `pytest` with the coverage plugin to avoid this.

PR #202, opened 2020-05-13
Remove redundant pytest

DylanBaker:


DylanBaker:
closes #201

.circleci/config.yml
```diff
@@ -54,16 +54,10 @@ jobs:
...
           command: |
             . venv/bin/activate
-            python -m pytest -p no:randomly --record-mode=none
-
-      - run:
-          name: coverage
-          command: |
-            . venv/bin/activate
-            coverage run --source spectacles -m pytest
+            coverage run --source spectacles -m pytest -p no:randomly --record-mode=none
```
joshtemple, line 60: I think it's to our benefit to run randomly because it catches issues where one test depends on the state of another. It reveals implicit dependencies between tests due to the ordering.

If we get a flaky test, randomly prints the seed used so it's straightforward to reproduce.
If we experien

DylanBaker: @joshtemple 👍 Just updated.

PR #203, opened 2020-05-13
Create context manager for branch management

joshtemple:
Plus a few fixes:
- Cleaning up logging messages
- Fixing odd test behavior with `parser` fixture in `test_cli.py`
- Complete refresh of tests cassettes (because of new branch behavior)


spectacles/cli.py
```diff
@@ -579,7 +594,7 @@ def iter_errors(lookml: List) -> Iterable:
             )
         if mode == "batch":
             logger.info(
-                f"\n\nTo determine the exact dimensions responsible for {'this error' if project.number_of_errors == 1 else 'these errors'}, "
+                f"\n\nTo determine the exact dimensions responsible for {'this error' if len(errors) == 1 else 'these errors'}, "
```
joshtemple, line 597: @DylanBaker, is this right?

DylanBaker:
@joshtemple I've also tested the import projects functionality locally and it seems to be doing its thing.

spectacles/client.py
```diff
@@ -146,14 +146,41 @@ def update_workspace(self, project: str, workspace: str) -> None:
...
+                title="Couldn't get all Git branches.",
+                status=response.status_code,
+                detail=(
+                    f"Unable to get all Git branches in project '{project}'. "
+                    "Please try again."
+                ),
+                response=response,
+            )
+
+        return [branch["name"] for branch in response.json()]
```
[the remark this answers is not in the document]
joshtemple: Nope, will change it to be more consistent!

spectacles/cli.py
joshtemple: Just wanted to make sure it wasn't an artifact of the rebases because I thought that `number_of_errors` was a new feature you added?

DylanBaker: Yeah. I'm not sure how I missed this in my work, but because we aren't returning the project anymore, we don't have the project to call it from. So we can now just count errors.

I think your work must have been happening in parallel and I just wasn't thinking ahead enough.

Issue #211, opened 2020-05-19
analytics-python needs to be added to setup.py


DylanBaker:
Current error when running 0.17.0rc1:
```
Traceback (most recent call last):
  File "/mnt/e/envs/ubuntu-python/bin/spectacles", line 5, in <module>
    from spectacles.cli import main
  File "/mnt/e/envs/ubuntu-python/lib/python3.7/site-packages/spectacles/cli.py", line 15, in <module>
    import spectacles.tracking as tracking
  File "/mnt/e/envs/ubuntu-python/lib/python3.7/site-packages/spectacles/tracking.py", line 1, in <module>
    import analytics  # type: ignore
ModuleNotFoundError: No module named 'analytics'
```

PR #212, opened 2020-05-19
Update dependencies in setup.py

joshtemple:


DylanBaker:
Should we change the circle config so that it doesn't reference requirements.txt? Otherwise, we aren't actually testing the actual pypi version?

joshtemple:
I don't think so, we want to pin dependencies for our test environment. If you find any evidence to the contrary though, let me know.

joshtemple:
Closes #211

PR #235, opened 2020-06-16
Allow model/explore selection and exclusion in assert

### LINKED ISSUE #165: Allow users to be able to run assert (data tests) on specific models
Currently, the `assert` command runs all data tests. It appears [from the docs](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/project#run_lookml_test) that we could specific which models it should run. We should allow that workflow.

It appears that you can't specify this at an explore level via the run API endpoint, but it's possible the explore is attached to the returned data in the 'Get All LookML' tests endpoint and that we could filter from there and specify exactly which tests we want to run.

### LINKED ISSUE #227: Assert validator should error if no tests are found
Thinking that maybe if we run `assert` and there are no tests that it should result in an error.

Reasoning is because the tests need to be included in a model or view order to be found and run, if they're unwittingly removed then they won't be run and if you're not paying attention you'll still get a nice green build for it.

joshtemple:
Closes #227, closes #165. Refactors the selection logic to use regex, making it much more concise and readable. Move the `--explores` and `--exclude` arguments to the validator subparser.

The architecture of this change means it's much more difficult to validate the selectors and exclusions passed in (for example, raising an error on `model_a/explore_c` to communicate that `explore_c` doesn't exist). This isn't ideal, because it's nice to catch when people have typos in their args, for example.

### DIFF
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -79,6 +79,12 @@ class DataTestValidator(Validator):
         selectors: Optional[List[str]] = None,
         exclusions: Optional[List[str]] = None,
     ) -> Dict[str, Any]:
+        # Assign default values for selectors and exclusions
+        if selectors is None:
+            selectors = ["*/*"]
+        if exclusions is None:
+            exclusions = []
+
         tests = self.client.all_lookml_tests(self.project)
 
         # The error objects don't contain the name of the explore
--- a/spectacles/validators.py
+++ b/spectacles/validators.py
@@ -98,6 +98,16 @@ class DataTestValidator(Validator):
                 test_to_explore[test["name"]] = test["explore_name"]
 
         test_count = len(selected_tests)
+        if test_count == 0:
+            raise SpectaclesException(
+                name="no-data-tests-found",
+                title="No data tests found.",
+                detail=(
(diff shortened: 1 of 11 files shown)


spectacles/validators.py
```diff
@@ -73,21 +74,55 @@ def __init__(self, client: LookerClient, project: str):
...
 
+        test_results: List[Dict[str, Any]] = []
+        for test in selected_tests:
+            test_name = test["name"]
+            model_name = test["model_name"]
+            results = self.client.run_lookml_test(
+                self.project, model=model_name, test=test_name
+            )
+            test_results.extend(results)
+
```
DylanBaker, line 123: I don't think there's anything we can do about this, but this will kick them off sequentially, right?

In theory, if a whole model is going to be tested, we could test it in one go I think, which would get us a decrease in run time because Looker would kick them off simultaneously. I think to do this we'd have to change the selection logic though.

joshtemple: Yeah, the most we could do with the endpoint would be to run them on a model level and then filter the results. Didn't seem worth the lift to me until there's demonstrated performance issues.

PR #236, opened 2020-06-16
Add content validator

### LINKED ISSUE #138: Add a content validator
Spectacles users would benefit from an additional validator that runs Looker's Content Validation via the API. There are a number of open questions about how this should work. This validator could expose the output in two possible ways:

* All content that requires validating on that branch. This would be a simple call to the content validation endpoint.
* All content that requires validating on that branch **that doesn't** require validating on the master branch, i.e. things that this branch actually broke. This would require two calls and comparison of content between the two. 

My gut feel is that we probably want to provide both ways, possibly controlled by a flag.

joshtemple:
Closes #138.

Adds a validator for content validation `spectacles content` with the following validator-specific parameters:

- `--incremental`: ignore errors that are already present in production
- `--exclude-personal`: ignore errors in content in personal folders

There are a couple other notable differences from the Looker content validator:

- Can use `--explores` and `--exclude` to test specific LookML for content issues
- Only returns errors for the project specified by `--project` instead of for the entire instance

Misc changes also included in this PR:
 - Validator module broken up into 3 validator-specific module and moved to its own subpackage
 - Print help message on query mode in dim text
 - Suggest that the project name is wrong if error encountered getting active Git branch
 - LookML objects now have a list of errors instead of a single error 
 - `log_duration` moved to the CLI functions instead of the Runner methods
 - Selector parsing moved to an independent module and now uses regex for easier code
 - Validators now print the first 6 chars of the commit or the branch reference being validated against


joshtemple:
This PR needs tests before it's ready for review.

joshtemple:
@DylanBaker, there are a few limitations to this validator in the current state:

1. If an explore is deleted, we won't be able to associate the content error to the `Project` because the errors returned by the API are not project-qualified. It is possible to fix this and display dangling errors (#240) but it will require additional dev because currently we prune the `Project` to only the models and explores that have been selected.
1. For the same reason, if there are two projects with the same model with the same explore in each, we could appropriate the error to the wrong project. This is a Looker API limitation, to my knowledge there is no way around this at the moment.

DylanBaker:
@joshtemple Thanks. That makes sense. It sounds like we could solve both those issues if we could convince Looker to return the project name from that endpoint?

joshtemple:
That's correct. I've created #240 and #241 to track this missing functionality and I'll make sure it's in the docs PR as caveats as well.

PR #330, opened 2021-03-03
Add query profiler for SQL validator

joshtemple:


spectacles/cli.py
```diff
@@ -519,6 +521,24 @@ def _build_sql_subparser(
         help="Specify how many concurrent queries you want to have running \
             against your data warehouse. The default is 10.",
     )
+    subparser.add_argument(
+        "-p",
+        "--profile",
+        action="store_true",
+        help=(
+            "After validation, display queries that took longer than 5 seconds to "
+            "complete."
```
DylanBaker, line 530: Do we want to exclude 5 from this description, given it's configurable?

DylanBaker (approved):
One small comment, but otherwise this looks good to me.

spectacles/cli.py
joshtemple: Definitely.

PR #345, opened 2021-03-24
v1.0.0rc2

### LINKED ISSUE #344: New commits are not available to be checked out
Creating a new branch from a commit ref is not working in some cases because that commit is not present in the local repo yet.

joshtemple:
Closes #344.

Improves VCR cassette response filtering so access tokens are not accidentally pushed by contributors.

Includes changes from `feature/limit-field` to trim down the info we request from the Looker API which should substantially speed up runs for larger Looker instances.


DylanBaker (approved):
LGTM.

Issue #274, opened 2020-07-24
Ignoring spectacles on dimensions that don't have sql parameter


leighajarett:
For example, location dimensions (have sql_lat and sql_lon) and duration dimensions (have sql_start and sql_end)

PR #370, opened 2021-06-22
Add the ability to ignore a dimension using the tags field

Ester-Ramos:
# Context

Some dimensions like `distance`, `location` or `duration` don't have the `sql` field available so there's no way to ignore these dimensions if you want to.

# Feature

This PR adds the possibility to ignore a dimension if you add the tag `"spectacles: ignore"` in the `tags` field.

According to the Looker documentation:

```
 These strings are not used by the Looker model, but can be passed to other applications.
```

so this feature won't interfere with the customers.

Closes https://github.com/spectacles-ci/spectacles/issues/274


joshtemple:
Thanks @Ester-Ramos, this is an awesome change! I wasn't aware of `tags`, which could be super useful for some other things we're working on. I'm happy to merge this change and we'll cut a new release soon.

PR #376, opened 2021-06-28
Exclude-folders and include flags on content validation

### LINKED ISSUE #373: Provide ability to filter content validator to include or exclude specific folders
Currently, we let users exclude errors from personal folders, but they can't get more granular than that. We should allow users to (1) run the content validator only on specific folders and (2) exclude specific folders from the content validator.

**Acceptance Tests**

- [ ] Can pass `--folders` with a comma separated list of integers. This would filter the content validation results for those folders
- [ ] Can pass `--exclude-folders` with a comma separated list of integers. This would filter out any content validation results for those folders.

vpipkt:
Closes #373


DylanBaker:
@vpipkt Thanks for this! Really appreciate you picking this up.

There's no way for you to have known this, given I didn't document it in the issue, but we'd like to include or exclude content from folders passed via the CLI _and_ all of their children folders. i.e. if folder 1 is passed to the CLI with the `--folders` flag, and it contains folders 2 and 3, we'd like to also include (and check for) folders 2 and 3 when we do the filtering. 

I think this could be achieved with a recursive function that gets the folder definition from the API and looks for the children over and over until there aren't any nested folders. We could then check against that larger list. This would require us using the [Get All Spaces endpoint](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/space#get_all_spaces).

That said, also very much open to other implementations. Do you have any thoughts here?

vpipkt:
@DylanBaker you're welcome! 

I think that seems reasonable behaviour to have children included. Do you think a similar approach would make sense for `--exclude-folders`? 

A use case that is important to my org is to be able to have looker developers exclude selected subfolders in their personal space.  So we might call with `--exclude-folders 99` where 99 is the folder `dev-user/scratch`  and expect also `dev-user/scratch/project_a` and `dev-user/scratch/foo` to be exluded.

Apologies that I have not attempted unit testing here. The test framework is reasonably complicated and I don't have access to the test looker instance. I'll keep working on it and  someone else can help with testing.

DylanBaker:
@vpipkt Yeah, we need to make this testing easier for external developers. I'm happy to pick that up!

And yeah, I think it should apply to both the include and exclude folders flag.

joshtemple:
@vpipkt, you can run `pytest -m "not vcr"` to exclude the tests that make live API calls. You should be able to run all of those tests without defining any credentials. Let me know if that's not the case!

spectacles/cli.py
```diff
@@ -577,6 +579,22 @@ def _build_content_subparser(
...
+        nargs="+",
+        help="Exclude errors found in folders specified by id.",
+        default=[],
+    )
+
+    subparser.add_argument(
+        "--folders",
+        type=int,
+        nargs="+",
+        help="Include errors found in folders specified by id, takes precedence over --exclue-personal or --exclude-folders.",
```
joshtemple, line 594: Typo in this line, also exclude should take precedence as we discussed @DylanBaker.

spectacles/validators/content.py
```diff
@@ -22,6 +31,32 @@ def _get_personal_folders(self) -> List[int]:
...
         return personal_folders
 
+    def _get_all_subfolders(self, folders: List[int]) -> List[int]:
+        result = []
+        for folder in folders:
+            result.extend(self._get_subfolders(folder))
+        return result
+
+    def _get_subfolders(self, folder_id: int) -> List[int]:
+        all_folders = self.client.all_folders(self.project.name)
```
joshtemple, line 41: Instead of getting all folders every time (which could get hefty), let's use the [Search Folder endpoint](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/folder#search_folders) where we can pass `parent_id` to get the IDs of the children!

spectacles/validators/content.py
```diff
@@ -22,6 +31,32 @@ def _get_personal_folders(self) -> List[int]:
...
+        subfolders = [folder_id]
+        children = [
+            child["id"] for child in all_folders if child["parent_id"] == folder_id
+        ]
+        if children:
+            for child in children:
+                subfolders.extend(self._get_subfolders(child))
+        return subfolders
+
+    def _ignore_folder_result(self, folder_id: Optional[str]) -> bool:
```
joshtemple, line 51: I like the `is_*` naming convention for bool-returning functions. What about `is_selected`? That would be consistent with the way we describe this sort of thing for explores and models, plus it makes the if statement cleaner in `validate` (no continue needed).

joshtemple (changes requested):

joshtemple:
Also, would like to see a bunch of tests added to make sure this logic works the way we think it does.

spectacles/validators/content.py
DylanBaker: What about getting all folders once and then passing them around? I'd rather reduce the number of API calls.

joshtemple: I think that's fine too. If it's a _huge_ tree of folders, it's possible that more Search calls could be faster than one Get All call, but I'm probably splitting hairs here.

DylanBaker:
@vpipkt I'm going to close this and replace it with #404. We had to disable running tests on forked branches a while back and haven't yet got around to putting in place the new CI setup. For the moment, it's easiest for me to checkout a new branch and open a new PR. Thanks for all your work here! We're going to get it merged in shortly.

vpipkt:
@DylanBaker Thanks so much for your help,#404 is going to make our builds better!!

PR #394, opened 2021-08-14
Change retry to apply to all client requests + for HTTPError

### LINKED ISSUE #379: Spectacles should retry errors that might be transient
Semi-regularly Looker seems to return transient 4xx and 5xx errors (particularly often around git branch operations). While we do handle the exceptions gracefully, it is likely worthwhile to try those requests a few times before having the entire run fail. 

Open question as to whether this should be specific API calls or all API calls.

DylanBaker:
## Change description

We find that the Looker API can semi-regularly provide transient errors. We already had some logic to retry after Timeout errors. This PR extends that retry to _all_ requests and now also retries after HTTPErrors.

## Type of change

## Related issues

Closes #379 

### Security

### Code review 

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -3,7 +3,7 @@ import time
 from dataclasses import dataclass
 import backoff  # type: ignore
 import requests
-from requests.exceptions import Timeout
+from requests.exceptions import Timeout, HTTPError
 import spectacles.utils as utils
 from spectacles.types import JsonDict
 from spectacles.logger import GLOBAL_LOGGER as logger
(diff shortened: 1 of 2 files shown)


joshtemple:
I wonder if it's better to take a slightly more gradual approach and only retry for 500 HTTP errors from Looker. This means we'll be unnecessarily retrying 400s and things that are our/the user's fault.

joshtemple (commented):
I wonder if it's better to take a slightly more gradual approach and only retry for 500 HTTP errors from Looker. This means we'll be unnecessarily retrying 400s and things that are our/the user's fault.

DylanBaker:
Part of the problem is that it often seems to be git related issues that are transient, which return 4XX and not 500s.

joshtemple:
As we discussed, preference here would be to only decorate the problematic endpoints to retry 4xx responses so we don't have to worry about retries potentially delaying in situations where the error is the client's fault.

joshtemple (approved):

PR #395, opened 2021-08-14
Reset Client to development mode if required after 60 minute timeout

### LINKED ISSUE #392: When API authentication is refreshed after one hour, we lose workspace state
After 60 minutes, Looker API creds expire and we refresh them. However, when we refresh the credentials, we end up back in the `production` workspace. Therefore, if the run is testing a branch or commit (1) we start running queries in the wrong workspace and (2) we can't reset the branch successfully at the end of the run. 

When we refresh the credentials, we should be workspace aware and reset to the right workspace.

DylanBaker:
## Change description

60 minutes after authenticating the client, we need to re-authenticate with the Looker client. When we do so, the workspace gets reset to production, even if we had previously been in development mode. This change fixes that behaviour. If we were in development mode when the token expires, we will reset to development mode after re-authenticating.

## Type of change

## Related issues

Closes #392 

### Security

### Code review 

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -80,6 +80,7 @@ class LookerClient:
         self.api_version: float = api_version
         self.access_token: Optional[AccessToken] = None
         self.session: requests.Session = requests.Session()
+        self.requires_dev_mode: bool = False
 
         self.authenticate()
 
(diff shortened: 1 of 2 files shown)


joshtemple (approved):
I pushed a commit to make this slightly more explicit and make it more simple (no dependency on `runner.py` now, the client tracks its own workspace). If you're happy with it, I'm happy with it.

PR #402, opened 2021-08-30
Add 'successes' to result of data test validator

### LINKED ISSUE #397: Return all data tests run

DylanBaker:
## Change description

We have had a request from a number of customers to return _all_ data tests, not just passing data tests. This PR achieves this by doing the following:

- Adds a `successes` key to the `Explore` class.
- Adds all passing data tests as JSON to the `successes` key on the explore.
- Formats the results for consumption by the hosted service

## Type of change

## Related issues

Closes #397 

### Security

### Code review 

### DIFF
--- a/spectacles/validators/data_test.py
+++ b/spectacles/validators/data_test.py
@@ -50,7 +50,12 @@ class DataTestValidator(Validator):
             model_name = test["model_name"]
             explore_name = test["explore_name"]
             query_url_params = test["query_url_params"]
-            file_path = test["file"].split("/", 1)[1]
+            
+            try:
+                file_path = test["file"].split("/", 1)[1]
+            except IndexError:
+                raise SpectaclesException(f"Couldn't extract file path from unexpected file '{test['file']}'")
+            
             explore_url = (
                 f"{self.client.base_url}/explore/{model_name}"
                 f"/{explore_name}?{query_url_params}"
--- a/spectacles/validators/data_test.py
+++ b/spectacles/validators/data_test.py
@@ -50,6 +50,16 @@ class DataTestValidator(Validator):
             model_name = test["model_name"]
             explore_name = test["explore_name"]
             query_url_params = test["query_url_params"]
+            file_path = test["file"].split("/", 1)[1]
+            explore_url = (
+                f"{self.client.base_url}/explore/{model_name}"
+                f"/{explore_name}?{query_url_params}"
+            )
+            lookml_url = (
+                f"{self.client.base_url}/projects/{self.project.name}"
+                f"/files/{file_path}?line={test['line']}"
+            )
+
             results = self.client.run_lookml_test(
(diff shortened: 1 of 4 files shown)


spectacles/validators/data_test.py
```diff
@@ -50,6 +50,16 @@ def validate(self) -> Dict[str, Any]:
             model_name = test["model_name"]
             explore_name = test["explore_name"]
             query_url_params = test["query_url_params"]
+            file_path = test["file"].split("/", 1)[1]
```
joshtemple, line 53: Is this reliable for all paths? Worried about KeyErrors happening here.

spectacles/validators/data_test.py
```diff
@@ -50,6 +50,16 @@ def validate(self) -> Dict[str, Any]:
...
+            explore_url = (
+                f"{self.client.base_url}/explore/{model_name}"
+                f"/{explore_name}?{query_url_params}"
+            )
```
joshtemple, line 57: I guess the earlier PR didn't actually implement this?

spectacles/validators/data_test.py
```diff
@@ -50,6 +50,16 @@ def validate(self) -> Dict[str, Any]:
...
+            lookml_url = (
+                f"{self.client.base_url}/projects/{self.project.name}"
+                f"/files/{file_path}?line={test['line']}"
+            )
```
joshtemple, line 61: Nice! Hadn't realized we had everything needed to return this.

joshtemple (commented):
Generally looks good. I would like to see a test or two before we merge. Noting for viewers that we're planning to refactor this work a bit more cleanly in a future PR, this is a short-term implementation.

spectacles/validators/data_test.py
DylanBaker: It did. I'm just moving it outside of the loop below as we already have all the info and need it for both the error and success path so it stops us from declaring it twice.

spectacles/validators/data_test.py
DylanBaker: I think you had actually done the first version of this. I'm just pulling it up from the later loop.

spectacles/validators/data_test.py
DylanBaker: Which bit are you worried about? I'm pretty confident the `file` key is on every test returned and it is always prefixed with `projectname/`.

DylanBaker:
@joshtemple I'll definitely add some more tests. Otherwise, questions answered in thread.

spectacles/validators/data_test.py
joshtemple: A completely unfounded concern that the path format might differ in different setups. I would be satisfied if we catch an `IndexError` when we do the `split` + indexing and return a sensible error message about the path having an unexpected format.

joshtemple (approved):

PR #403, opened 2021-08-30
Add tile type and title to content validation errors

### LINKED ISSUE #374: Provide name of dashboard tile in content validator when relevant
Currently we only provide the name of the dashboard in the content validator. When we link to a dashboard with many tiles, it can be hard to find exactly where the error is. We should provide the name of the tile.

**Acceptance Criteria**

- [ ] Data output from assert validator should include tile when relevant
- [ ] CLI should print the tile name when relevant.

DylanBaker:
## Change description

When we raise a content validation error for a dashboard, it can be hard to know where the error is coming from in the dashboard as there can be many tiles and filters. This PR adds two new fields to the content validation error, one specifying whether it is a visualisation or filter, and another specifying the title of the visualisation or filter. This should make it easier to pinpoint where content validation errors are coming from. 

## Type of change

## Related issues

Closes #374  

### Security

### Code review 

### DIFF
--- a/spectacles/validators/content.py
+++ b/spectacles/validators/content.py
@@ -46,7 +46,6 @@ class ContentValidator(Validator):
             else:
                 self._handle_content_result(content, content_type)
 
-        print(self.project.get_results(validator="content"))
         return self.project.get_results(validator="content")
 
     @staticmethod
--- a/spectacles/validators/content.py
+++ b/spectacles/validators/content.py
@@ -46,6 +46,7 @@ class ContentValidator(Validator):
             else:
                 self._handle_content_result(content, content_type)
 
+        print(self.project.get_results(validator="content"))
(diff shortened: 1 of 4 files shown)


spectacles/validators/content.py
```diff
@@ -46,6 +46,7 @@ def validate(self) -> Dict[str, Any]:
             else:
                 self._handle_content_result(content, content_type)
 
+        print(self.project.get_results(validator="content"))
```
joshtemple, line 49: Guessing this needs to be deleted.

joshtemple (changes requested):

spectacles/validators/content.py
DylanBaker: Yes. It does. Good catch.

joshtemple (approved):

PR #404, opened 2021-08-30
Add folder inclusion and exclusion for content validator

### LINKED ISSUE #373: Provide ability to filter content validator to include or exclude specific folders

DylanBaker:
## Change description

This PR is a replacement for PR #376. It is being re-created so that tests can be run in CircleCI. We haven't yet got things configured properly for forked branches to be run. 

This PR adds the ability to run the content validator on specific folders or to exclude specific folders from the result. It adds two new arguments: `--folders` and `--exclude-folders`. The exclusion list will take precedence over the inclusion list, i.e. if folder `1` is both included and excluded, it won't be selected. 

## Type of change

## Related issues

Closes #373 

### Security

### Code review 


spectacles/validators/content.py
```diff
@@ -22,6 +32,31 @@ def _get_personal_folders(self) -> List[int]:
                 personal_folders.append(folder["id"])
         return personal_folders
 
+    def _get_all_subfolders(self, input_folders: List[int]) -> List[int]:
+        result = []
+        all_folders = self.client.all_folders(self.project.name)
```
joshtemple, line 37: Looks like we still call this twice. Are you okay with that?

tests/test_content_validator.py
```diff
@@ -83,3 +83,79 @@ def test_personal_folder_content_should_not_be_present(self, validator_fail):
...
+            yield validator, results
+
+    def test_personal_folder_content_should_not_be_present(
+        self, validator_fail_with_exclude
+    ):
+        results = validator_fail_with_exclude[1]
+        assert len(results["errors"]) == 0
+
+
+class TestValidateFailIncludeFolder:
```
joshtemple, line 113: Can you add a test for what happens when the user specifies a folder ID that doesn't exist?

DylanBaker: Yes, I'll add that. Good call.

spectacles/validators/content.py
DylanBaker: I think I am. It's now once for personal folders and once for subfolders, which I think I'm happy with. Would you be inclined to consolidate further?

joshtemple: Only if it's straightforward to do so. I don't think 1 extra call is going to have much of a performance hit.

joshtemple (approved):
LGTM!

PR #463, opened 2021-11-23
Attach model, explore and test names to data tests with SQL errors

DylanBaker:
## Change description

When a SQL error occurs during a data test, we don't attach the model name, explore name and test name to the error object because the JSON response from Looker is different. This PR makes a small change to fix this.

## Type of change

### Security

### Code review 


joshtemple:
Does it make sense to raise an actual SqlError instead, or do we not have the right info to populate those fields?

DylanBaker:
We don't have the right info unfortunately.

joshtemple (approved):

joshtemple:
LGTM!

PR #477, opened 2021-12-28
Add incremental SQL validation (lite)

### LINKED ISSUE #369: Enable incremental SQL validation
We would like to enable users to specify a source and target branch for incremental SQL validation. During incremental validation, Spectacles should only test the dimensions where the generated SQL for the dimension's query is different on the source branch than on the target branch.

For example, if the SQL for the dimension `duration` is the following on `develop` branch:

```sql
SELECT ROUND(time / 60, 2) AS duration FROM times
```

...but the SQL for the dimension `duration` is the following on `main` branch:

```sql
SELECT ROUND(time, 2) as duration FROM times
```

...Spectacles *should* test the `duration` dimension.

Acceptance criteria:
 - [ ] Can pass `--incremental`
 - [ ] `--incremental` with no following args tests source against target == production
 - [ ] Source branch defined as presence of absence of `branch` or `commit`
 - [ ] Only validates dimensions found in the diff

### LINKED ISSUE #371: Support incremental content validation with non-prod targets
Currently we support incremental content validation comparing the source branch (defined by either `--branch` or `--commit`) vs. production. As a Spectacles user, I may want to content validate the diff of my changes between my source branch and a non-production target branch (e.g. `develop` or `qa`).

Acceptance criteria:

- [ ] Can pass `--incremental`
- [ ] Passing `--incremental` with no arguments is the current behavior—testing vs. production
- [ ] Can add an argument after `--incremental` (branch name or commit) that cause Spectacles to perform content validation on the diff between the source branch and target branch/commit as specified by this argument

joshtemple:
### Adding incremental validation to SQL validator

This change implements incremental (diff) validation for the SQL validator. When running Spectacles on a branch or commit, user can provide the flag `--incremental` to compare their chosen branch/commit to production, and only test explores that have changed (i.e. compile to different SQL) vs. their branch/commit.

Furthermore, when testing at the dimension level (previously called hybrid or single mode), Spectacles will only show errors where the compiled SQL for that dimension differs when compared to production.

Currently, because compiling SQL for every dimension in two different Git contexts is prohibitive to performance, Spectacles runs all dimensions from explores that have different SQL, then checks the SQL of all errored dimensions against production and only returns the errors where the SQL is different. Future changes will implement improvements to allow us to compile SQL for each dimension in advance, which should enable even faster incremental validation.

If the user would prefer to compare to another branch or commit instead of production, they can use the `--target` argument to specify what Spectacles should compare to. For example,

```
spectacles sql --incremental --branch dev-josh --target dev-dylan
```

would only run SQL validation on explores that compile to different SQL on `dev-josh` than on `dev-dylan`.

### Changes to mode [Breaking]

To more closely match the Spectacles cloud app, we've removed the `--mode` argument, which previously allowed users to run in `batch`, `hybrid`, and `single` mode. Ultimately, there aren't many cases for running in single mode, since hybrid mode is faster and more efficient. For clarity, this changes the argument to `--fail-fast`, where running without `--fail-fast` (the default) is the same as hybrid mode, and running with `--fail-fast` is the same as batch mode.

### Changes to include/exclude syntax [Breaking]

This change also implements a simplification for specifying model/explores or content folders to include or exclude. Previously, users would have to use two separate arguments like `--explores` and `--exclude` or `--include-folders` and `--exclude-folders`. This change consolidates those arguments into `--explores` and `--folders`, where prepending a hyphen `-` instructs Spectacles to exclude the folder or explore. For example, `--explores -*/order_items` would exclude all explores named `order_items` from validation.

### Other changes

This change also improves how the branch manager handles ephemeral branches, some refactoring around building the project and validator structure, and a sizeable rewrite of the SQL validator and runner methods to support the incremental use case.

## Type of change

## Related issues

Closes #369, closes #371

### Security

### Code review 

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 ## Change description
 
-> Description here
+Description here
 
 ## Type of change
 - [ ] Bug fix (fixes an issue)
@@ -8,7 +8,7 @@
 
 ## Related issues
 
-> Closes [#1]() 
+Closes [#1]() 
 
 ## Checklists
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,4 +13,4 @@ repos:
     rev: v0.910
     hooks:
     - id: mypy
-      args: [--install-types, --non-interactive]
+      args: [--ignore-missing-imports]
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -862,35 +862,34 @@ def run_sql(
         runtime_threshold,
     )
 
-    explores = sorted(results["explores"], key=lambda x: (x["model"], x["explore"]))
-    for explore in explores:
-        message = f"{explore['model']}.{explore['explore']}"
-        printer.print_validation_result(passed=explore["passed"], source=message)
+    for test in sorted(results["tested"], key=lambda x: (x["model"], x["explore"])):
+        message = f"{test['model']}.{test['explore']}"
+        printer.print_validation_result(passed=test["passed"], source=message)
 
-    count_errors = 0
-    for explore in explores:
-        if not explore["passed"]:
-            for test in explore["tests"]:
-                for error in test["errors"]:
-                    count_errors += 1
-                    printer.print_sql_error(
-                        model=error["model"],
-                        explore=error["explore"],
-                        message=error["message"],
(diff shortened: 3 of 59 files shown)


DylanBaker (approved):
drake-lets-go

PR #482, opened 2022-01-05
Add-ons to LookML validation

joshtemple:
## Change description

 - Updates JSON schema to include LookML validator response definition
 - Adds `severity` argument for configuring
 - Cosmetic changes to printing to make validator output match the other validators more closely
 - Add line numbers to printed header to match IDE

## Type of change

## Related issues
N/A

### Security

### Code review 

### DIFF
--- a/spectacles/cli.py
+++ b/spectacles/cli.py
@@ -661,15 +661,23 @@ def run_lookml(
         commit_ref,
     )
 
-    errors = sorted(results["errors"], key=lambda x: (x["metadata"]["file_path"]))
+    errors = sorted(results["errors"], key=lambda x: x["metadata"]["file_path"] or "a")
+    unique_files = sorted(
+        set(
+            error["metadata"]["file_path"]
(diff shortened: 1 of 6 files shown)


DylanBaker (commented):
All looks good to me, except for one thing:

The severity flag is restricted to the exit code in the CLI, but doesn't affect the JSON that is returned from the runner/validator. I think this is fine for the CLI, but in the context of the web application, the 'status' might therefore not be correct. 

i.e. someone sets up a suite with severity threshold of error. If the run returns warnings but no errors, the status will be 'failed', when I think we really want it to pass. We could add some logic to the application to handle this, but it feels cleaner to do it in the CLI.

My suggestion would be to pass the severity threshold to the runner/validator and conditionally set the status based on the results returned. The `raise_error` logic in the CLIP then becomes simpler because we just look at the status of the validator.

DylanBaker (approved):

PR #491, opened 2022-01-17
Allow users to ignore hidden fields in SQL validation

### LINKED ISSUE #163: Feature Request: ignore hidden: yes columns
Just validated a large model with lots of hidden fields that are hidden because they aren't supposed to be used on thier own. I'm aware of `-- spectacles: ignore` and the discussion in #35 but I think there could be a --ignore-hidden that doesn't test hidden columns that kills two birds with one stone

Cheers - thanks for your awesome work!

DylanBaker:
## Change description

Testing hidden fields can be redundant. They can't be exposed to the user and if they are referenced in another dimension, the SQL will be tested as part of the SQL of the downstream field.

I should note, I don't suggest we ignore hidden fields by default. Sometimes users will temporarily unhide a field to add it to a report, and then re-hide the field. The field (and its SQL) are therefore still being used and we should test them by default.

## Type of change

## Related issues

Closes #163 

### Security

### Code review 

### DIFF
--- a/spectacles/lookml.py
+++ b/spectacles/lookml.py
@@ -20,6 +20,7 @@ class Dimension(LookMlObject):
         type: str,
         tags: List[str],
         sql: str,
+        is_hidden: bool,
         url: Optional[str] = None,
     ):
         self.name = name
@@ -29,6 +30,7 @@ class Dimension(LookMlObject):
         self.tags = tags
         self.sql = sql
         self.url = url
+        self.is_hidden = is_hidden
         self.queried: bool = False
         self.errors: List[ValidationError] = []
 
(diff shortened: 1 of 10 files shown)


spectacles/lookml.py
```diff
@@ -429,6 +433,10 @@ def build_project(
         if include_dimensions:
             for explore in model.explores:
                 explore.dimensions = build_dimensions(client, model.name, explore.name)
+                if ignore_hidden_fields:
+                    explore.dimensions = [
+                        dim for dim in explore.dimensions if not dim.is_hidden
+                    ]
```
joshtemple, line 439: Should we move this logic into `build_dimensions`? That's already where we filter out `ignore` so it's probably more consistent.

joshtemple (approved):
This is looking good. My only ask is to verify this works in as expected in incremental cases. What happens if a dimension is hidden on target but not hidden on base or vice-versa?

PR #502, opened 2022-02-08
Run explore-level queries in chunks

### LINKED ISSUE #474: Spectacles queries can hit character limits on BigQuery
BigQuery has a 1024 character limit for SQL queries. This means that the initial 'batch' query can sometimes be too large. 

There is no way in the current implementation of Spectacles to get around this. We will need to find some way to 'chunk' the queries.

Things to answer:

* How should we chunk them? By view? By an arbitrary number? (view could be helpful because it reduces joins per query)
* Should we consider the chunking in which dimension level queries we kick off? i.e. if only one chunk fails, should we only run those dimensions?
* Dose the implementation of this need to take the incremental implentation into account in any way?

joshtemple:
## Change description

This change breaks up explore-level queries into chunks of N dimensions, where N is configured by a constant in the SQL validator module. Previously, the atomic unit of SQL validation was the `SqlTest`, which had a 1:1 relationship with a query ID and query task.

Now, a `SqlTest` has a one-to-many relationship with query IDs and their subsequent query task. Note that for the purpose of the `explore_url` and `sql`, each `SqlTest` **still uses a single query with all dimensions selected, regardless of the chunk size.** This allows us to continue to do incremental validation without any changes.

This change also removes the instance variable `_running_tests` which isn't actually necessary, since you can use the `_test_by_task_id` dictionary for the same purpose.

Because this fundamentally alters the requests made by the SQL validator, I've overwritten the cassettes for `test_runner` and `test_sql_validator`.

## Type of change

## Related issues

Closes #474.

### Security

### Code review 


spectacles/validators/sql.py
```diff
@@ -169,11 +178,23 @@ def _create_explore_test(
...
+        # Create separate chunked queries for execution, we don't store compiled SQL
+        # or the Explore URL for these queries
+        chunk_queries: List[Query] = []
+        for chunk in chunks(dimensions, size=CHUNK_SIZE):
+            query = self.client.create_query(
+                explore.model_name, explore.name, chunk, fields=["id", "share_url"]
+            )
+            chunk_queries.append(Query(query["id"]))
+
+        sql = self.client.run_query(query["id"]) if compile_sql else None
```
DylanBaker, line 191: This client function feels like it could be named a bit better, given it doesn't actually run the query, as far as I can tell.

spectacles/validators/sql.py
```diff
@@ -1,17 +1,37 @@
...
+import itertools
 import time
+from spectacles.utils import chunks
 from spectacles.client import LookerClient
 from spectacles.lookml import Dimension, Explore, Project
 from spectacles.exceptions import SpectaclesException, SqlError
 from spectacles.logger import GLOBAL_LOGGER as logger
 from spectacles.printer import print_header
 
+CHUNK_SIZE = 10
```
DylanBaker, line 13: Should we make this dynamic? This feels like overkill for a really small explore, and maybe not enough for a large one.

DylanBaker (commented):
This looks good to me, though I'm going to do some manual testing now too.

What happens in fail-fast if multiple chunks fail? Do we return all the errors?

Am I right in thinking that this doesn't actually currently prune the dimension level queries? i.e. if one chunk fails, we query all dimensions?

spectacles/validators/sql.py
DylanBaker: Oh. I've actually misunderstood this. We run in chunks of 10, not 10 chunks. Right?

DylanBaker:
> What happens in fail-fast if multiple chunks fail? Do we return all the errors?

Answering my own question: yes, we show multiple chunks. 

I'm honestly not sure what the right behaviour here should be.

DylanBaker:
It seems the first chunk that runs has 11 dimensions instead of 10. All subsequent ones have 10 or fewer. Not sure why that's happening.

spectacles/validators/sql.py
joshtemple: Chunks of 10, correct. I'm not at all attached to this number. It should probably be way larger. I'm not sure if we need to make it configurable but we probably should tune it.

spectacles/validators/sql.py
joshtemple: I've kept the names of the client methods 1:1 with the Looker API endpoints for simplicity. You do actually use the Run Query endpoint with the `sql` format to compile.

DylanBaker (approved):

PR #516, opened 2022-02-24
Properly handle failed runs without dimension errors

### LINKED ISSUE #512: Carry through explore-level error when no dimension level errors
Occasionally, Spectacles identifies an error with the explore-level query but doesn't find any errors in the dimension-level queries. This results in a a failing state with no errors that is confusing to the user.

This can happen for two reasons: (1) something about the underlying database (or code if testing prod) changes during the run and as a result the explore fails on the first pass but then works fine when we get to the dimensions, or (2) there’s an ambiguous column reference error which only occurs when we test all the dimensions together, but doesn’t occur at the dimension level check.

We should check for this situation and propagate the explore-level error(s).

joshtemple:
## Change description

This change fixes an issue where the explore-level query would error but the dimension-level queries would pass, resulting in a run status inconsistent with error results.

This change adds some logic to the way results are retrieved from the project to return the explore-level errors if no dimension-level errors are found. However, this naive approach broke the way incremental errors are handled, since we couldn't differentiate between the incremental case and the problematic case.

So I've introduced a new attribute on the ValidationError class, `ignore`, which is used by incremental runs to tell the `get_results` method to skip those errors as duplicated.

## Type of change

## Related issues

Closes #512

### Security

### Code review 

### DIFF
--- a/spectacles/runner.py
+++ b/spectacles/runner.py
@@ -349,12 +349,12 @@ class Runner:
                 )
 
                 for dimension in project.iter_dimensions(errored=True):
-                    dimension.errors = [
-                        error
-                        for error in dimension.errors
-                        if not isinstance(error, SqlError)
-                        or (dimension.name, error.metadata["sql"]) not in target_sql
-                    ]
+                    for error in dimension.errors:
+                        if (
+                            isinstance(error, SqlError)
+                            and (dimension.name, error.metadata["sql"]) in target_sql
+                        ):
+                            error.ignore = True
 
         results = project.get_results(validator="sql", fail_fast=fail_fast)
         return results
--- a/spectacles/exceptions.py
+++ b/spectacles/exceptions.py
@@ -79,6 +79,7 @@ class ValidationError(GenericValidationError):
         self.model = model
         self.explore = explore
         self.metadata = metadata
+        self._ignore: bool = False
         super().__init__()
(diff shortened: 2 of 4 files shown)


spectacles/runner.py
```diff
@@ -349,12 +349,12 @@ def validate_sql(
...
-                        for error in dimension.errors
-                        if not isinstance(error, SqlError)
-                        or (dimension.name, error.metadata["sql"]) not in target_sql
-                    ]
+                    for error in dimension.errors:
+                        if (
+                            isinstance(error, SqlError)
+                            and (dimension.name, error.metadata["sql"]) in target_sql
+                        ):
+                            error.ignore = True
```
DylanBaker, line 357: @joshtemple Can you talk me through what's happening here? I realised I'm not sure I follow the logic (even pre-change). Thanks!

joshtemple: The idea is to ignore errors from dimensions with SQL that matches the SQL of dimensions on the target ref (incremental skips). Previously, we just deleted those errors entirely. Now, we keep them but mark them as ignored.

DylanBaker (approved):

joshtemple:
Closing because we decided to merge this change before merging `feature/chunk-queries` (#502), instead merging via #517.

PR #519, opened 2022-03-01
Provide ability for users to pin locally imported projects to specific git refs

### LINKED ISSUE #454: Users should be able to set branches for upstream projects
Currently, if a user is testing a project that imports another project, we 'pin' the upstream project to production.

However, there are some cases when you're changing the upstream project and want to understand what impact it will have on the downstream project before changes are made to the upstream project. In this circumstance, you want to be able to run Spectacles on the downstream project (where you would find errors) but using a branch of the upstream project.

This needs refining, but my initial proposal is that there is a new CLI argument that takes `project: ref` key-value pairs.

DylanBaker:
## Change description

This PR updates the Runner, LookerBranchManager and CLI to allow users to run Spectacles with locally imported Looker projects pinned to specific git refs.

i.e. if the project `eye_exam` currently locally imports `welcome_to_looker`, the prior behaviour would have always tested `welcome_to_looker` @ production. Now, by passing the following CLI arg, we can test it at any git ref:

```bash
--import-refs welcome_to_looker:some-branch-name
```

## Type of change

## Related issues

Closes #454 

### Security

### Code review 


spectacles/runner.py
```diff
@@ -33,16 +33,24 @@ def is_commit(string: str) -> bool:
...
         self.client = client
         self.project = project
         self.remote_reset = remote_reset
+        self.import_refs = import_refs
 
         state: ProjectState = self.get_project_state()
         self.workspace: str = state.workspace
         self.history: List[ProjectState] = [state]
+        # TODO: Should this call happen after we've checked out the correct state?
         self.imports: List[str] = self.get_project_imports()
```
DylanBaker, line 54: @joshtemple Is it possible this call is happening too early? We do it when we instantiate the LookerBranchManager, which is typically on production at that point. 

If someone is working on a branch where they have added a local import, we won't catch it when we call it then. 

I think it's possible that we need to make this call _after_ we've landed in the target ref state.

DylanBaker:
@joshtemple These tests are passing when I run with `--record-mode=new_episodes` but not after that when I try running with `--record-mode=none`. I'm not quite sure what I am doing wrong. Would you be able to have a look?

spectacles/cli.py
```diff
@@ -489,6 +499,13 @@ def _build_validator_subparser(
             In order to test a specific commit, Spectacles will create a new branch \
             for the tests and then delete the branch when it is finished.",
     )
+    group.add_argument(
+        "--import-refs",
+        nargs="+",
+        default=[],
+        help="The git ref at which any locally imported Looker projects should be set to. \
+            List of strings in project_name:ref_name format. ",
```
joshtemple, line 507:
```suggestion
        help="Pin locally imported Looker projects to a specific ref (Git branch or commit) during validation. \
            Provide these arguments in project_name:ref format.",
```

spectacles/cli.py
```diff
@@ -209,6 +209,10 @@ def preprocess_dashes(args: List[str]) -> List[str]:
     return processed
 
 
+def process_import_refs(input: List[str]) -> dict:
+    return {n.split(":")[0]: n.split(":")[1] for n in input}
```
joshtemple, line 213:
```suggestion
    return dict(arg.split(":") for arg in input)
```

joshtemple: We should have some error handling here in case the format is wrong. Try/catch on the right exception.

joshtemple (commented):
This is pretty slick. I'm really happy to see how straightforward this was with the branch manager. Have a few small suggestions. I'll take a look at the cassette stuff.

PR #524, opened 2022-03-02
Handle additional severity levels

joshtemple:
## Change description

Handle other LookML validator severity options from the Looker API like `fatal` and `success`.

Description here

## Type of change

### Security

### Code review 

### DIFF
--- a/spectacles/printer.py
+++ b/spectacles/printer.py
@@ -93,11 +93,11 @@ def print_lookml_error(
 ) -> None:
     if file_path is None:
         file_path = "[File name not given by Looker]"
-    header_color = yellow if severity == "warning" else red
+    header_color = red if severity in ("fatal", "error") else yellow
     print_header(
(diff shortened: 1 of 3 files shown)


spectacles/printer.py
```diff
@@ -93,11 +93,11 @@ def print_lookml_error(
 ) -> None:
     if file_path is None:
         file_path = "[File name not given by Looker]"
-    header_color = yellow if severity == "warning" else red
+    header_color = red if severity in ("fatal", "error") else yellow
```
Hawk94, line 96: We don't need to account for info / success here, right?

Hawk94 (approved):
LGTM

spectacles/printer.py
joshtemple: They'll just show up yellow. I tried blues but it didn't look right.

PR #544, opened 2022-03-31
Check for cached LookML validation results before running full validator

DylanBaker:
### Using cached LookML validation results

Some users are experiencing time-out issues with the LookML validator. We can try to used the cached results first. If those are not present or are stale, we will then run the full validator.

This change also add some significant bug fixes and improvements:

### Compare SqlTests for equality at a less granular level

Previously, we were comparing SqlTests for equality using the entire LookML objects associated with the test (e.g. Explore). Because the LookML URLs for dimensions can change across branches, this meant that tests for an Explore with equivalent SQL on the target and base branches were not being marked as equivalent. This meant we were testing Explores in incremental mode that we didn't need to.

This change makes the comparison less specific, only on the basis of explore name, model name, and compiled SQL.

### Improve logging printing for readability

I've added some indentation and separation to branch manager logs so that they're easier to read in debug mode.

### Refresh all cassettes and prefer pytest-recording

I've deleted and re-created all cassettes and used pytest-recording `pytest.mark.vcr` everywhere instead of VCR.py context managers. This improves readability and also allows us to mandate that Authorization headers are filtered by default.

### Secret checking

I've added a pre-commit hook and CI check to look for tokens accidentally committed in VCR.py cassettes and some instructions in CONTRIBUTING.md.

## Type of change

### Security

### Code review 


spectacles/client.py
```diff
@@ -840,6 +840,29 @@ def lookml_validation(self, project) -> JsonDict:
...
+
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            raise LookerApiError(
+                name="unable-to-get-cached-lookml-validation",
+                title=f"Couldn't get cached LookML validation results in project {project}.",
+                status=response.status_code,
+                detail=(
+                    "Failed to get cached LookML valiation results. Please try again."
```
joshtemple, line 856: If there are any error cases we do understand, would be good to elaborate on them in the error message, like "Cache results may have expired, please re-run the LookML validator to populate the cache."

spectacles/client.py
```diff
@@ -840,6 +840,29 @@ def lookml_validation(self, project) -> JsonDict:
...
 
+    def cached_lookml_validation(self, project) -> Optional[JsonDict]:
+        logger.debug(f"Getting cached LookML validation results for '{project}'")
+        url = utils.compose_url(self.api_url, path=["projects", project, "validate"])
+        response = self.get(url=url, timeout=TIMEOUT_SEC)
+
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            raise LookerApiError(
```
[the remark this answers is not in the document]
DylanBaker: If there's no cached result, we get a 204 No Content. My proposal in this scenario is that we return a `None` from the API and handle that accordingly in the validator.

spectacles/client.py
DylanBaker: There aren't any cases that I'm aware of that cause this error to be triggered. If there is no cached result, it returns a 204 instead of a 4XX.

tests/cassettes/init_client.yaml
```diff
@@ -5153,4 +5153,246 @@ interactions:
     status:
       code: 200
       message: OK
+- request:
```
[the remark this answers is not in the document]
DylanBaker: File removed.

joshtemple (approved):

spectacles/printer.py
```diff
@@ -79,9 +79,19 @@ def print_content_error(
 
 
 def print_data_test_error(
-    model: str, explore: str, test_name: str, message: str, lookml_url: str
+    model: Optional[str],
+    explore: Optional[str],
+    test_name: Optional[str],
```
DylanBaker, line 84: What are the cases where these aren't defined?

joshtemple: This seems to happen when there's a SQL error in a data test's explore.

joshtemple: Not in the case I was seeing! But maybe something was failing further upstream?

PR #555, opened 2022-04-18
Handle 404 when compiling SQL

### LINKED ISSUE #554: Getting compiled SQL for explores that don't exist on branch/commit generate an unhandled error
When running the SQL validator incrementally, if any explore fails, we run tests for all dimensions, then get the compiled SQL from the target branch for failed dimensions so we can filter them out if they are actually unchanged. 

Often, the errors are for explores or dimensions that don't exist in the target branch. 

If it's a case of a dimension not existing on the target, but the dimension is within an explore that _does_ exist, we get a 200 status with the following text: 
```
-- SQL could not be generated because of errors with this query.
```

However, if the errors are on a new explore (i.e. one that doesn't yet exist on the target), we get a 404 status with a much longer text error message. 

This causes Spectacles to fail with an error of the following format:
```
Couldn't retrieve compiled SQL.
Failed to retrieve compiled SQL for query '11771'. Please try again.
```

We need to better handle these 404s such that we can reliably do this comparing for new explores/models that don't yet exist on the target.

DylanBaker:
## Change description

This changes allows handling for the `run_query` client function for when a 404 is returned. This can be because an explore or model doesn't exist on the branch and we need to return a string for incremental diff comparison.

## Type of change

## Related issues

Closes #554 

### Security

### Code review 

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -903,17 +903,22 @@ class LookerClient:
         response = self.get(url=url, timeout=TIMEOUT_SEC)
         try:
             response.raise_for_status()
-        except requests.exceptions.HTTPError:
-            raise LookerApiError(
-                name="unable-to-retrieve-compiled-sql",
-                title="Couldn't retrieve compiled SQL.",
-                status=response.status_code,
-                detail=(
-                    f"Failed to retrieve compiled SQL for query '{query_id}'. "
-                    "Please try again."
-                ),
-                response=response,
-            )
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 404:
+                return (
+                    "-- SQL could not be generated because of errors with this query."
+                )
+            else:
+                raise LookerApiError(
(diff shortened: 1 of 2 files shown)


joshtemple (approved):

kurnito365:
Hi @DylanBaker, should the issue we saw in our call thereotically be fixed with this merged PR?

DylanBaker:
Hey @kurnito365. We're going to merged this into the application this afternoon, at which point yes that should be resolved for you.

kurnito365:
> Hey @kurnito365. We're going to merged this into the application this afternoon, at which point yes that should be resolved for you.

Which afternoon? I think we are in different timezones 😅

DylanBaker:
Ha. European afternoon. This is now deployed, so you should be good to go.

kurnito365:
> Ha. European afternoon. This is now deployed, so you should be good to go.

Thanks Dylan!

PR #572, opened 2022-05-20
Report "dangling" content validator errors

### LINKED ISSUE #240: Return dangling content validation errors
It's possible that the content validator could return content errors that are no longer associated with an existing explore. When this happens, Spectacles could display the errors anyway or display a message asking the user to run validation without selecting or excluding anything to see the errors.

It would be helpful to reproduce the API response in this case and decide how to handle this edge case.

DylanBaker:
## Change description

We currently prune the LookML tree when building the project. This means that content validator errors relating to explores that don't exist anymore don't get reported by Spectacles. This PR changes this functionality so that these errors now appear.

The changes work in the following way:

* When we build the project for content validation runs, we now return the entire LookML tree. This allows us to attach all errors to their relevant place on the tree, even if the explore has been excluded from the run. This is what allows us to distinguish between explores that have been deleted and explores that have been excluded by the filters.
* When we attach the errors to the LookML tree, we first check if the relevant explore exists. If it does, we attach the error there. We then check if the relevant model exists. If it does, we attach the error there. This second part is new. Previously, we were only attaching to existing explores.
* When we return the errors for the project, we now allow for filters to be passed to the `get_results` function. This will only happen during the content validator. When this occurs, we prune the errors to just the relevant LookML tree.

This changes mean that we will now return all content validation errors relevant to a project, except for those that are linked to models that don't exist anymore. There is no way for us to know where to attach those, so they will remain dangling. 

## Type of change

## Related issues

Closes #240 

### Security

### Code review 

### DIFF
--- a/spectacles/lookml.py
+++ b/spectacles/lookml.py
@@ -339,23 +339,20 @@ class Project(LookMlObject):
             # We create an explore "tested" record for those that
             # aren't in the LookML tree.
             distinct_explores = set()
-            if filters is None and model.errors:
-                # Raise name error if there are errors on the model and filters
-                # isn't being passed. We need filters when running the content
-                # validator, which is the only time a model will have errors.
-                raise TypeError(
-                    "The argument 'filters' is required in Content Validator."
-                )
-            elif filters is not None and model.errors:  # appease mypy
-                for error in model.errors:
-                    if is_selected(model.name, error.explore, filters):
-                        distinct_explores.add(error.explore)
-                        errors.append(error.to_dict())
-                model_tested = [
-                    {"model": model.name, "explore": e, "status": "failed"}
-                    for e in distinct_explores
-                ]
-                tested.extend(model_tested)
+
+            for error in model.errors:
+                if filters is not None and not is_selected(
+                    model.name, error.explore, filters
+                ):
+                    continue
+                distinct_explores.add(error.explore)
+                errors.append(error.to_dict())
+
+            model_tested = [
+                {"model": model.name, "explore": e, "status": "failed"}
+                for e in distinct_explores
+            ]
+            tested.extend(model_tested)
 
             for explore in model.explores:
                 if filters is not None and not is_selected(
--- a/spectacles/lookml.py
+++ b/spectacles/lookml.py
@@ -343,17 +343,14 @@ class Project(LookMlObject):
                 # Raise name error if there are errors on the model and filters
                 # isn't being passed. We need filters when running the content
                 # validator, which is the only time a model will have errors.
-                raise NameError(
+                raise TypeError(
                     "The argument 'filters' is required in Content Validator."
                 )
             elif filters is not None and model.errors:  # appease mypy
-                for error in [
-                    e
-                    for e in model.errors
-                    if is_selected(model.name, e.explore, filters)
-                ]:
-                    distinct_explores.add(error.explore)
-                    errors.append(error.to_dict())
+                for error in model.errors:
+                    if is_selected(model.name, error.explore, filters):
+                        distinct_explores.add(error.explore)
+                        errors.append(error.to_dict())
                 model_tested = [
                     {"model": model.name, "explore": e, "status": "failed"}
                     for e in distinct_explores
@@ -460,7 +457,7 @@ def build_project(
             ),
         )
 
-    # For the content validator, we need all explores.
+    # Prune to selected explores for non-content validators
     if not include_all_explores:
         for model in models:
             model.explores = [
@@ -469,13 +466,17 @@ def build_project(
                 if is_selected(model.name, explore.name, filters)
             ]
 
-            if include_dimensions:
-                for explore in model.explores:
-                    explore.dimensions = build_dimensions(
-                        client, model.name, explore.name, ignore_hidden_fields
-                    )
-        project = Project(name, [m for m in models if len(m.explores) > 0])
-        return project
-    else:
+    if include_dimensions:
+        for model in models:
+            for explore in model.explores:
+                explore.dimensions = build_dimensions(
+                    client, model.name, explore.name, ignore_hidden_fields
+                )
+
+    # Include empty models when including all explores
+    if include_all_explores:
         project = Project(name, models)
-        return project
+    else:
+        project = Project(name, [m for m in models if len(m.explores) > 0])
+
+    return project
--- a/spectacles/lookml.py
+++ b/spectacles/lookml.py
@@ -474,6 +474,8 @@ def build_project(
                     explore.dimensions = build_dimensions(
                         client, model.name, explore.name, ignore_hidden_fields
                     )
-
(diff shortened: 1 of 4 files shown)


spectacles/lookml.py
```diff
@@ -424,18 +460,22 @@ def build_project(
...
-        if include_dimensions:
-            for explore in model.explores:
-                explore.dimensions = build_dimensions(
-                    client, model.name, explore.name, ignore_hidden_fields
-                )
-
-    project = Project(name, [model for model in models if len(model.explores) > 0])
-    return project
+    # For the content validator, we need all explores.
+    if not include_all_explores:
```
joshtemple, line 464: I think this would be slightly easier to read if we swapped the if statement to `if include_all_explores`.

spectacles/lookml.py
```diff
@@ -424,18 +460,22 @@ def build_project(
...
+        for model in models:
+            model.explores = [
+                explore
+                for explore in model.explores
+                if is_selected(model.name, explore.name, filters)
+            ]
+
+            if include_dimensions:
```
joshtemple, line 472: This assumes that we only ever use `include_dimension` if `include_all_explores` is False. That's true today but do you think (to prevent this from biting us if that assumption changes) we should shift the logic so it's possible to include dimensions while also including all explores?

spectacles/lookml.py
```diff
@@ -322,14 +325,46 @@ def get_explore(self, model: str, name: str) -> Optional[Explore]:
...
         for model in self.models:
+            # Add model level content validation errors.
+            # We create an explore "tested" record for those that
+            # aren't in the LookML tree.
+            distinct_explores = set()
+            if filters is None and model.errors:
+                # Raise name error if there are errors on the model and filters
+                # isn't being passed. We need filters when running the content
+                # validator, which is the only time a model will have errors.
+                raise NameError(
```
joshtemple, line 346: I usually see `NameError` reserved for [code-related issues](https://docs.python.org/3/library/exceptions.html#NameError), not runtime stuff like this one. I would prefer raising a `TypeError` here, since that's the problem (`filters` can't be of type `None`).

spectacles/lookml.py
```diff
@@ -322,14 +325,46 @@ def get_explore(self, model: str, name: str) -> Optional[Explore]:
...
+                    "The argument 'filters' is required in Content Validator."
```
joshtemple, line 347: Also, the more I think about this, from the standpoint of `get_results`, I don't see a reason why you couldn't ask for results with model errors and no filters. I think probably it's better for us to validate this a layer up when it's called in the Content Validator itself rather than here.

spectacles/lookml.py
```diff
@@ -322,14 +325,46 @@ def get_explore(self, model: str, name: str) -> Optional[Explore]:
...
+                )
+            elif filters is not None and model.errors:  # appease mypy
+                for error in [
+                    e
+                    for e in model.errors
+                    if is_selected(model.name, e.explore, filters)
+                ]:
```
joshtemple, line 354: Nit: Generally not a fan of looping over a conditional list comprehension like this. I think it's more straightforward to write a plain `for` loop with a nested `if`, that way you only read one `for` instead of two.

spectacles/lookml.py
```diff
@@ -322,14 +325,46 @@ def get_explore(self, model: str, name: str) -> Optional[Explore]:
...
+                    distinct_explores.add(error.explore)
+                    errors.append(error.to_dict())
+                model_tested = [
+                    {"model": model.name, "explore": e, "status": "failed"}
+                    for e in distinct_explores
```
joshtemple, line 359: Doesn't this end up creating duplicate tested records with the records that are created for the explore-level errors? Or would this be completely distinct because those errors couldn't be associated with an explore?

spectacles/validators/content.py
```diff
@@ -131,8 +131,9 @@ def _get_errors_from_result(
             explore: Optional[Explore] = project.get_explore(
                 model=model_name, name=explore_name
             )
+            model: Optional[Model] = project.get_model(model_name)
```
joshtemple, line 134: I think we can make this a little faster by calling `get_model` on the Project first, then calling `get_explore` on the returned Model instead of on the Project, that will cut down on extra iteration.

spectacles/validators/content.py
```diff
@@ -131,8 +131,9 @@ def _get_errors_from_result(
...
             # Skip errors that are not associated with selected explores
```
joshtemple, line 135:
```suggestion
            # Skip errors that are not associated with selected explores or existing models
```

spectacles/validators/content.py
```diff
@@ -154,8 +155,11 @@ def _get_errors_from_result(
                         else None
                     ),
                 )
-                if content_error not in explore.errors:
+                if explore and content_error not in explore.errors:
                     explore.errors.append(content_error)
                     content_errors.append(content_error)
+                elif model and content_error not in model.errors:
+                    model.errors.append(content_error)
+                    content_errors.append(content_error)
```
joshtemple, line 163: This feels easier to read in a nested if.

```suggestion
                if content_error not in explore.errors:
                    if explore:
                        explore.errors.append(content_error)
                        content_errors.append(content_error)
                    else:
                        model.errors.append(content_error)
                        content_errors.append(content_error)
```

joshtemple (commented):
Thanks for your work on this! I have some feedback below. Also, regarding tests: have you added any tests that explicitly test for this situation? That would be good to do in addition to refactoring existing ones.

spectacles/lookml.py
DylanBaker: Yeah, I'll move the logic out.

spectacles/lookml.py
DylanBaker: I think there's possibly a few issues with that approach, but I could be wrong:

* Even if in the content validator we only attached 'selected' errors to the tree, we would still be returning the whole tree. Then, when we call `get_results`, you would end up with records in `tested` for explores that should not be tested as far as the end-user is concerned. 
* If we wanted resolve the issue in the point above, we would need to prune the tree in the content validator, which doesn't really feel like something that is in it's domain, but if we're happy to do that, we could definitely move it there.

spectacles/validators/content.py
DylanBaker: Agreed. I'll make this change.

spectacles/lookml.py
```diff
@@ -410,7 +440,7 @@ def build_project(
     fields = ["name", "project_name", "explores"]
     for lookmlmodel in client.get_lookml_models(fields=fields):
```
joshtemple, line 441: I'm not sure how important this is, but we loop through models 4 times (441, 459, 467, 477). Do you think we can consolidate some of these loops? I don't consider it blocking for the PR but if you have some time it might give us a bit of efficiency and readability.

joshtemple (approved):

PR #586, opened 2022-06-28
Async requests and binary search

### LINKED ISSUE #147: Binary Search
I've got some rather large explores (900+ dimensions) where single mode runs quite slowly simply due to the _number_ of queries run. A good way to cut down on that might be to implement an algorithm like the following:

1. Run a batch query on the whole explore
2. If it passes, do nothing
3. If it fails, split the explore in half and repeat steps 1-2 for each half.

If you've got a massive explore with few broken dimensions, this could save a lot of queries!

### LINKED ISSUE #575: Upgrade default Looker API version to 4.0
Looker API 4.0 is now in General Availability, which means that deprecation of v3.1 will eventually become an issue. We should upgrade Spectacles to use Looker API v4.0, following [the migration guides linked here](https://developers.looker.com/api/advanced-usage/version-4-ga).

joshtemple:
## Change description

This change is a major performance improvement for Spectacles, especially for the SQL validator. Spectacles makes a large volume of API requests to Looker. Moving to asynchronous requests allows Spectacles to issue requests much faster, resulting in faster runtimes.

This change also adds a binary search approach to identifying SQL errors in an Explore. Previously, when testing an Explore with N dimensions, Spectacles would run 1 query in fail fast mode (the Explore-level query) and N + 1 queries otherwise (one Explore-level query and N dimension-level queries).

Now, Spectacles uses binary search, bisecting the Explore into 2 queries that each cover N / 2 dimensions. If either query fails, Spectacles bisects it again into equally sized queries. If either query passes, Spectacles stops querying there.

In cases where users have many Explores with many dimensions but few errors (i.e. a team that has cleaned up their project of SQL errors but is testing new PRs), binary search will be _much_ faster. In our testing, we observed speedups 2-8x faster than the current version.

### Pydantic

I've incorporated the library Pydantic as a first step to replacing jsonschema in this library. Pydantic is a better approach to runtime type checking because it uses type hints and native classes (called models). Eventually, I'd like to entirely remove jsonschema and use Pydantic for checking the Project results, but for now I've used it to type the Looker API's query result response in the SQL validator. We can land and expand from there over time.

### Changes to tests

 * Async/await in Python can lead to coroutine creep, which is what happened here. As such I've had to refactor the entire test suite to handle the necessary `await` statements.
 * I've removed VCR testing entirely. Through the test suite refactor, I realized that VCR wasn't even working because of the way we'd structured fixtures. I tried to make it work, but the library is more trouble than it's worth. Instead, I've broken out integration tests (which make live API calls to our test Looker project) and unit tests (which are fully mocked) into their own CI steps and project folders. This will allow us to run unit tests even on forked PRs from contributors, while saving integration tests for manual approval.
 * I updated the CircleCI image in our config to their latest generation, and fixed some issues that arose from that upgrade

### Other miscellaneous changes

 * This change drops support for Looker API 3.1 and has some necessary migrations to 4.0, namely IDs like folder IDs and query IDs are no longer int but str type. It also removes the `space` metadata from ContentErrors, which is no longer returned by the API.
 * I've added a `py.typed` marker so that Python code that uses Spectacles can benefit from type checking
 * This version no longer supports Python 3.7. Users will have to upgrade to Python 3.8 or greater.
 * Switched from `requests` to `httpx`. I liked `httpx`'s async HTTP client the best and it no longer made sense to keep requests just for the handful of synchronous requests when `httpx` can do the same. I've also used the excellent companion library, `respx` for mocking HTTP calls.

## Type of change

## Related issues

Closes #147, closes #575

### Security

### Code review 


joshtemple:
@DylanBaker could I get an approving review on this when you get a chance? Would like to merge and start a release candidate.

DylanBaker (approved):

DylanBaker:
@joshtemple, do we have all the recent non-async changes incorporated in this PR?

PR #673, opened 2023-01-03
remove redundant variable

### LINKED ISSUE #671: Data tests can cause "user query limit" issue if queries queue for over 600 seconds.
With the recent async changes, we now run _all_ data tests at the same time. 

This is from [the Looker docs](https://cloud.google.com/looker/docs/admin-panel-database-queries):

> Per-user query limit and timeout: To prevent any single user from filling up the Looker query queue, each user has a maximum number of allowed concurrent queries and a corresponding queue timeout. By default, each user can run a maximum of 15 concurrent queries, and the timeout for queries queued due to this limit is 600 seconds.

Some organisaitons have a large number of tests, many of which take a while to run. As a result of this, some organisations are getting errors related to the user query limit when the data test queries take a long time. 

I believe the solution is to bound the async tasks with a Semaphore. It probably makes sense to make that number 15, though we could also make it configurable.

DylanBaker:
## Change description

With the recent async changes, we now run _all_ data tests at the same time. 

This is from [the Looker docs](https://cloud.google.com/looker/docs/admin-panel-database-queries):

> Per-user query limit and timeout: To prevent any single user from filling up the Looker query queue, each user has a maximum number of allowed concurrent queries and a corresponding queue timeout. By default, each user can run a maximum of 15 concurrent queries, and the timeout for queries queued due to this limit is 600 seconds.

Some organisaitons have a large number of tests, many of which take a while to run. As a result of this, some organisations are getting errors related to the user query limit when the data test queries take a long time. 

This PR sets a concurrency limit of 15 to the data tests.

## Type of change

## Related issues

Closes #671 

### Security

### Code review 

### DIFF
--- a/spectacles/validators/data_test.py
+++ b/spectacles/validators/data_test.py
@@ -93,40 +93,44 @@ class DataTestValidator:
 
     async def validate(self, tests: List[DataTest]) -> List[DataTestError]:
         data_test_errors: List[DataTestError] = []
+        query_slot = asyncio.Semaphore(
+            15
+        )  # This is the per-user query limit in Looker for most instances
 
-        async def run_test(test: DataTest) -> None:
-            results = await self.client.run_lookml_test(
-                test.project_name, model=test.explore.model_name, test=test.name
-            )
-            test.explore.queried = True
-            result = results[0]  # For a single test, list with length 1
-
-            if result["success"]:
-                test.passed = True
-                test.explore.successes.append(
-                    {
-                        "model": test.explore.model_name,
-                        "explore": test.explore.name,
-                        "metadata": {
-                            "test_name": result["test_name"],
-                            "lookml_url": test.lookml_url,
-                            "explore_url": test.explore_url,
-                        },
-                    }
+        async def run_test(test: DataTest, query_slot: asyncio.Semaphore) -> None:
+            async with query_slot:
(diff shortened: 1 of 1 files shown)


spectacles/validators/data_test.py
```diff
@@ -93,40 +93,44 @@ async def get_tests(self, project: Project) -> List[DataTest]:
...
-                        "model": test.explore.model_name,
-                        "explore": test.explore.name,
-                        "metadata": {
-                            "test_name": result["test_name"],
-                            "lookml_url": test.lookml_url,
-                            "explore_url": test.explore_url,
-                        },
-                    }
+        query_slot = asyncio.Semaphore(
+            15
```
joshtemple, line 97: Let's define this `15` as a constant at the top of the module.

joshtemple (commented):
This looks good!

I wonder, should we set the SQL validator's query concurrency to 15 by default as well? If Looker is capping the per-user query concurrency at 15, I'm not sure it makes sense to allow users to set concurrency any higher, or even to configure that value. From my testing with async, I struggled to get 20+ queries running simultaneously anyway.

spectacles/validators/data_test.py
DylanBaker: Will do!

joshtemple (approved):

PR #705, opened 2023-04-13
Validate only changed LookML

joshtemple:
## Change description

This change leverages an undocumented Looker endpoint to perform partial LookML validation, which is much faster than validating the entire project every time. This is the same endpoint used internally by the Looker browser IDE.

We've included a fallback to the full validation 

## Type of change

### Security

### Code review 

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -907,6 +907,32 @@ class LookerClient:
         return result
 
     @backoff.on_exception(backoff.expo, BACKOFF_EXCEPTIONS, max_tries=DEFAULT_MAX_TRIES)
+    async def new_lookml_validation(self, project) -> Optional[JsonDict]:
+        logger.debug(
+            f"Getting LookML validation results for altered files for '{project}'"
+        )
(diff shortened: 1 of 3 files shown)


joshtemple:
Waiting to merge this one until we have a bit more context from Looker about how this endpoint works. It's currently failing an integration we wouldn't expect it to fail: returning a passing status for a very old untouched branch that definitely has LookML errors, which makes us wonder if it's interacting with a cache in a way that we don't expect.

joshtemple:
We've learned from Looker that the new endpoint unfortunately doesn't achieve parity with the IDE. It lacks AST validation (this is performed client-side in the Looker IDE). This means it will miss certain errors that would otherwise be caught by using `/validate`, e.g. errors in `test` or `set`.

We will continue to encourage Looker to support an API-based solution that has runtime and validation parity with the IDE and are continuing the conversation with their product team. Fingers crossed we have something workable shortly! 🤞

I will close this PR for now until we have a workable solution to discuss.

PR #716, opened 2023-05-23
Wait for expired queries to resolve themselves or retry

### LINKED ISSUE #111: Query results incorrectly assigned expired status by Looker API
Looker's async `multi_results` [endpoint](https://docs.looker.com/reference/api-and-integration/api-reference/v3.1/query#get_multiple_async_query_results) occasionally reports the status of a running query as `expired` right before the query completes successfully. This causes spectacles to incorrectly mark the query as errored, when in fact, the query was successful.

In a [previous change](https://github.com/spectacles-ci/spectacles/commit/0b463a11d151ef067f98cc98ddffe6115c7a2e66#diff-cbee2ab0268f8f001868af4ca066ba04L234-R237), we considered marking all `expired` queries as `complete`, but this becomes a problem if a query is killed manually (see report in #104) since it's subsequently marked as `expired`.

We will reach out to Looker to see if we can reproduce and fix the `expired` issue.

joshtemple:
## Change description

The Looker API returns the status `expired` for certain queries. However, sometimes it's a false alarm as the query eventually resolves itself to a completed status.

This change waits up to 30 seconds for that resolution to happen. If the query's status is still `expired` at that point, it puts the query back on the queue to be retried. After 2 retries it gives up and adds an error to the Explore.

## Type of change

## Related issues

Closes #111.

### Security

### Code review 

### DIFF
--- a/spectacles/validators/sql.py
+++ b/spectacles/validators/sql.py
@@ -422,11 +422,11 @@ class SqlValidator:
                         expired_for = time.time() - query.expired_at
                         if expired_for > EXPIRED_QUERY_WAIT_TIME:
                             # Stop waiting for query, decide if we should retry
-                            if query.expired_retries <= EXPIRED_RETRY_LIMIT:
+                            if query.expired_retries < EXPIRED_RETRY_LIMIT:
                                 logger.debug(
                                     f"Query task {task_id} expired for "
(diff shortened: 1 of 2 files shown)


joshtemple:
This PR still needs tests. I also need to investigate if a run with both Explore and dimension-level errors will be problematic for printing out in the CLI.

DylanBaker (approved):

PR #793, opened 2024-04-08
Use personal branches instead of temp branches when specified

DylanBaker:
## Change description

Use personal branches instead of temp branches when specified

## Type of change

### Security

### Code review 


spectacles/cli.py
```diff
@@ -529,6 +533,19 @@ def _build_validator_subparser(
         env_var="LOOKER_GIT_BRANCH",
         help="The branch of your project that Spectacles will use to run queries.",
     )
+    base_subparser.add_argument(
+        "--use-personal-branch",
+        action=EnvVarStoreTrueAction,
+        env_var="SPECTACLES_USE_PERSONAL_BRANCH",
```
joshtemple, line 539: Out of curiosity, why support as an environment variable?

spectacles/runner.py
```diff
@@ -126,7 +137,10 @@ async def __aenter__(self) -> None:
             self.branch = prod_state.branch
             self.commit = prod_state.commit
             if self.ephemeral:
-                self.branch = await self.checkout_temp_branch(prod_state.commit)
+                if self.use_personal_branch:
```
joshtemple, line 140: Instead of having to use this if/else every time, it would be tidier to have one function that checks out a "temp" branch, with a parameter that controls whether or not it's the personal branch.

spectacles/runner.py
```diff
@@ -249,6 +264,24 @@ async def get_project_imports(self) -> List[str]:
...
+        await self.client.hard_reset_branch(self.project, self.personal_branch, ref)
+        return self.personal_branch
+
+    async def get_personal_branch(self) -> str:
+        """Finds the name of the user's personal branch."""
+        branches = await self.client.get_all_branches(self.project)
+        for branch in branches:
+            if branch["personal"] and not branch["readonly"]:
+                return str(branch["name"])
+        raise ValueError("Personal branch not found")
```
joshtemple, line 283: Is there any other info we should raise here that might help with debugging? Maybe project?

tests/integration/test_runner.py
```diff
@@ -80,15 +81,21 @@ async def test_validate_data_tests_should_work(
     assert len(result["errors"]) > 0
 
 
+@pytest.mark.parametrize("use_personal_branch", [True, False])
```
joshtemple, line 84: Great way to test this!

spectacles/cli.py
DylanBaker: I honestly thought we supported all arguments as environment variables. I now realise we don't. Happy to remove it, but seems useful enough in a CI context to be able to set it as an environment variable once rather than on each validator command.

joshtemple: Okay, fine by me.

spectacles/runner.py
```diff
@@ -108,15 +111,17 @@ async def __aenter__(self) -> None:
         if self.branch:
             await self.update_workspace("dev")
             if self.ephemeral:
-                self.branch = await self.checkout_temp_branch("origin/" + self.branch)
+                await self.checkout_ephemeral_branch(
+                    "origin/" + self.branch, self.use_personal_branch, False
```
joshtemple, line 115: I'd probably do the flags as keyword args here since it's not obvious what `False` corresponds to.

spectacles/runner.py
```diff
@@ -267,6 +296,18 @@ async def checkout_temp_branch(self, ref: str) -> str:
...
+    ) -> None:
+        """Either check out temp or personal branch and hard-reset."""
+        if use_personal_branch:
+            if update_self_branch:
+                self.branch = await self.checkout_personal_branch(ref)
+            else:
+                await self.checkout_personal_branch(ref)
+        else:
+            self.branch = await self.checkout_temp_branch(ref)
+
```
joshtemple, line 310: My suggestion would be to return the branch and use that to update `self.branch` externally where necessary instead of having this function update `self.branch` on its own and needing the flag and extra conditional logic.

joshtemple (approved):

PR #803, opened 2024-05-10
Extend HTTP timeout

joshtemple:
## Change description

We were not backing off and retrying on manifest changes, which was by design since often the manifest is not there and we don't want to retry a 404. However, we can still retry timeouts, so I've added backoff for that endpoint for timeouts but not status errors.

We were not backing off and retrying on `httpx.ReadError`, so I've used the parent class `NetworkError` to catch this in our backoff logic.

We were only retrying most requests 3 times by default. With exponential backoff, this isn't waiting _that_ long in total. I've increased the number of retries to 10 for any network error or 502 status. I've ensured we give up immediately and don't retry on status error that isn't a 502.

## Type of change

### Security

### Code review 

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -81,13 +81,13 @@ files = [
 
 [[package]]
 name = "bandit"
-version = "1.7.7"
+version = "1.7.8"
 description = "Security oriented static analyser for python code."
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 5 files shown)


DylanBaker (commented):
@joshtemple doesn't this get over-ridden by the timeout on each request?

joshtemple:
Yes, and I'm still not sure why I'm seeing a 5 sec `ReadTimeout` in the logs. Thoughts on merging the backoff change for `TimeoutException` on the manifest call but removing the timeout changes?

DylanBaker:
Yeah, I'm good with that.

DylanBaker (approved):

PR #831, opened 2024-12-02
Devoted sql validation skip build project

jsnb-devoted:
## Change description

Description here

## Type of change

## Related issues

Closes [#1]() 

### Security

- [ ] Security impact of change has been considered
- [ ] Code follows security best practices and guidelines

### Code review 

- [ ] Pull request has a descriptive title and context useful to a reviewer

### DIFF
--- a/spectacles/client.py
+++ b/spectacles/client.py
@@ -66,17 +66,26 @@ class AccessToken:
         return False if time.time() < self.expires_at else True
 
 
+def log_backoff(details: dict) -> None:
+    logger.debug(
+        f"Backing off {details['wait']:0.1f} seconds after {details['tries']} tries. "
+        f"Error: {details['exception'].__class__.__name__}"
+    )
+
+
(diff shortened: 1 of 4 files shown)


DylanBaker (commented):
Hey @jsnb-devoted. Thanks for splitting this out.

While I totally understand how this makes sense for your use case, I think it probably leads to some complications on our end, particularly for how this works in the hosted application. 

Any run of Spectacles is typically scoped to a single project (via the `--project` flag). In a world where we don't build the project and just take explores directly from the filters, you could conceivably be testing explores from multiple different projects (whether they import each other or not). This behaviour isn't necessarily problematic in and of itself, but it would probably need to come with a broader rethinking of how we scope each validator at the project level. I'd be open to exploring this, but I think I'd need @joshtemple's input and we would probably need to work at it a little. Happy to do so, but for transparency we likely won't have time to work on it for at least a few weeks.

PR #854, opened 2025-08-19
Add read timeout handling with backoff and retry in Looker client

yanivbenzvi:
## Change description

Added `ReadTimeout` exception to the backoff retry mechanism in the Looker API client. This ensures that when HTTP read timeouts occur during API calls, the client will automatically retry the request using the existing exponential backoff strategy, improving reliability when dealing with slow or intermittent network connections.

## Type of change
- [ ] Bug fix (fixes an issue)
- [x] New feature (adds functionality)

## Related issues

[853](https://github.com/spectacles-ci/spectacles/issues/853) 

### Security

- [x] Security impact of change has been considered
- [x] Code follows security best practices and guidelines

### Code review 

- [x] Pull request has a descriptive title and context useful to a reviewer

PR #856, opened 2025-08-20
Read error fix

DylanBaker:
See #854


Hawk94 (approved):

DylanBaker:
@yanivbenzvi this is now merged in, so you could update your scripts to pip install directly from the right commit on `master`. Otherwise I'll be doing a release later this week.

